PR #31, opened 2019-04-15
service replica example

slimsag:


slimsag:
Support for this was merged, see https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/32

PR #37, opened 2019-08-19
Update zoekt for 3.7 release

keegancsmith:
This includes the final changes we want to ship for zoekt in 3.7. It requires
frontend to also be on the latest version of 3.7 since we made a minor change
to the zoekt API.

Part of https://github.com/sourcegraph/sourcegraph/pull/5290


slimsag:
Thanks for doing it here, too!

slimsag:
> It requires frontend to also be on the latest version of 3.7 since we made a minor change to the zoekt API.

I'll make sure this happens.

PR #38, opened 2019-08-19
add docker-compose yaml file

ggilmore:
TODO cc @slimsag:

- [x] Test grafana, etc. to see if they work correctly
- [x] All of the volumes assume that they'll be created under the current user's home directory. There should be an easier way to call this out/tweak this. Script? Environment variables (you can template these, see: https://docs.docker.com/compose/environment-variables/)?
- [x] This `docker-compose.yaml` file isn't portable since it has references to files relative to the root of the repository (grafana, nginx). Is this okay?
- [ ] Docs need to be written to describe how to scale each of these services (what to copy-paste, what environment variables to tweak, etc.). 
- [x] Do we need default [cpu-shares/quota, etc.](https://docs.docker.com/config/containers/resource_constraints/) settings for each service?


ggilmore:
cc @slimsag Right now, this is a straight translation of all the raw docker commands that `./deploy.sh` runs piped through https://bucherfa.github.io/dcc-web/

slimsag:
Note: There have been some recent changes in master (including some I am about to push tonight) that will need to be updated in this PR.

ggilmore:
@slimsag I just updated this off the latest master commit

docker-compose.yaml
```diff
@@ -0,0 +1,306 @@
...
+      - 'REPLACER_URL=http://replacer:3185'
+      - 'LSIF_SERVER_URL=http://lsif-server:3186'
+      - 'GRAFANA_SERVER_URL=http://grafana:3000'
+    volumes:
+      - >-
+        ~/sourcegraph-docker/sourcegraph-frontend-internal-0-disk:/mnt/cache
+    networks:
+      - sourcegraph
+    restart: always
+  sourcegraph-frontend-2:
```
ggilmore: I assume that we'll want to get rid of the multiple frontend containers from the default version of this config?

slimsag: We should investigate if we can use https://docs.docker.com/compose/compose-file/#replicas for this

docker-compose.yaml
```diff
@@ -0,0 +1,306 @@
...
+      - 'GRAFANA_SERVER_URL=http://grafana:3370'
+    volumes:
+      - >-
+        ~/sourcegraph-docker/sourcegraph-frontend-2-disk:/mnt/cache
+    ports:
+      - '0.0.0.0:3082:3080'
+    networks:
+      - sourcegraph
+    restart: always
+  sourcegraph-frontend-1:
```
slimsag: The docs we have in the scripts are rather helpful in trying to read these things -- as it stands this compose file is pretty illegible. We can just copy those docs into here.

docker-compose.yaml
```diff
@@ -0,0 +1,306 @@
...
+        ~/sourcegraph-docker/sourcegraph-frontend-1-disk:/mnt/cache
+    ports:
+      - '0.0.0.0:3081:3080'
+    networks:
+      - sourcegraph
+    restart: always
+  sourcegraph-frontend-0:
+    container_name: sourcegraph-frontend-0
+    image: >-
+      sourcegraph/frontend:3.10.4@sha256:5791fb16a1a2828f14c227160104eba3a251c876359e0cb33c48f832e3d2a53f
```
slimsag: I don't think(?) these need to be written as `>-` -- quoted strings like the env vars seems more clear to me?

docker-compose.yaml
ggilmore: @slimsag That feature is only supported for `docker swarm`, and it seems really unclear what the future of swarm is: https://github.com/docker/swarmkit/issues/2665

Regardless, I don't have personal experience with it. I can explore it more, but I'm hesitant to recommend it since its status seems to be in limbo.

docker-compose.yaml
ggilmore: Copied and re-ordered related services.

slimsag:
> All of the volumes assume that they'll be created under the current user's home directory. There should be an easier way to call this out/tweak this. Script? Environment variables (you can template these, see: https://docs.docker.com/compose/environment-variables/)?

Env vars seems best to me.

> This docker-compose.yaml file isn't portable since it has references to files relative to the root of the repository (grafana, nginx). Is this okay?

This seems fine for now. We can improve this later.

> Docs need to be written to describe how to scale each of these services (what to copy-paste, what environment variables to tweak, etc.).

We do not need this in order to merge this PR, since docker-compose is currently only aiming to replace sourcegraph/server which has no scaling. We can improve this later.

> Do we need default cpu-shares/quota, etc. settings for each service?

No.

docker-compose.yaml
```diff
@@ -0,0 +1,419 @@
+version: '2.4'
+services:
+  nginx:
```
slimsag: Add description and comments like other services have.

docker-compose.yaml
```diff
@@ -0,0 +1,419 @@
...
+    environment:
+      - GOMAXPROCS=8
+      - 'HOSTNAME=zoekt-webserver-0:6070'
+      - 'SRC_FRONTEND_INTERNAL=http://sourcegraph-frontend-internal:3090'
+    volumes:
+      - '~/sourcegraph-docker/zoekt-0-shared-disk:/data/index'
+    networks:
+      - sourcegraph
+    restart: always
+    hostname: zoekt-indexserver-0
```
slimsag:
```suggestion
    hostname: zoekt-indexserver-0

```

docker-compose.yaml
```diff
@@ -0,0 +1,419 @@
...
+  # Liveness probe: HTTP GET http://syntect-server:9238/health
+  # Ports exposed to other Sourcegraph services: 9238/TCP 6060/TCP
+  # Ports exposed to the public internet: none
+  #
+  syntect-server:
+    container_name: syntect-server
+    image: 'index.docker.io/sourcegraph/syntect_server:2b5a3fb@sha256:ef5529cafdc68d5a21edea472ee8ad966878b173044aa5c3db93bc3d84765b1f'
+    networks:
+      - sourcegraph
+    restart: always
```
slimsag:
```suggestion
    restart: always

```

docker-compose.yaml
```diff
@@ -0,0 +1,419 @@
...
+  # Ports exposed to the public internet: none
+  #
+  redis-store:
+    container_name: redis-store
+    image: 'index.docker.io/sourcegraph/redis-store:19-04-16_6891de821@sha256:56426d601ce1f6d63088fea1cefa61f69a2e809c7d90fc1d157cca63cf81b277'
+    volumes:
+      - '~/sourcegraph-docker/redis-store-disk:/redis-data'
+    networks:
+      - sourcegraph
+    restart: always
```
slimsag: This file is missing jaeger services. They should be added and commented out, so that we at least keep things in sync with one-another.

nginx/nginx/sourcegraph_server.conf
```diff
@@ -0,0 +1,4 @@
+# This config is generated by Sourcegraph. Do not modify, it will be rewritten
+# on startup. This config is included in the server context by nginx.conf
```
slimsag: Is this actually generated?

nginx/nginx.conf
```diff
@@ -0,0 +1,51 @@
+# You can adjust the configuration to add additional TLS or HTTP features.
```
slimsag: If you copied these nginx configs from server, please state so in a comment and indicate they need to be kept in sync with the upstream location. Otherwise, if they differ, that is fine as-is.

docker-compose.yaml
```diff
@@ -0,0 +1,419 @@
...
+  # Network: 100mbps
+  # Liveness probe: HTTP GET http://sourcegraph-frontend:3080/healthz
+  # Ports exposed to other Sourcegraph services: 6060/TCP
+  # Ports exposed to the public internet: 3080 (HTTP) and/or 3443 (HTTPS)
+  #
+  # Note: SRC_GIT_SERVERS, SEARCHER_URL, and SYMBOLS_URL are space-separated
+  # lists which each allow you to specify more container instances for scaling
+  # purposes. Be sure to also apply such a change here to the frontend-internal
+  # service.
+  sourcegraph-frontend-0:
```
slimsag: All of the services in this file need to have the `--cpus` and `--memory` copied over from the sh scripts

ggilmore: @slimsag Note, adding the `cpus` flags means that I can no longer run docker-compose up on my machine since it only has two cores.

ggilmore:
@slimsag PTAL

docker-compose.yaml
slimsag: This is expected; if you have less cores than is required you'll need to lower the CPU counts yourself. The docker-compose file should be optimized for a prod deployment setting initially I think

docker-compose/.env
```diff
@@ -0,0 +1,7 @@
+# This file contains default values for some environment variables that are used 
+# in docker-compose.yaml
+# See https://docs.docker.com/compose/environment-variables/ for more information.
+
+# The root storage location for various caches, gitserver, redis, postgres, etc.
+# Override this value to change where Sourcegraph stores this data.
+SOURCEGRAPH_HOME=~/sourcegraph-docker
```
slimsag: Suggest renaming this to  `DATA_ROOT`

docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,562 @@
...
+  # replicas.
+  nginx:
+    container_name: nginx
+    image: 'index.docker.io/library/nginx:1.17.7@sha256:8aa7f6a9585d908a63e5e418dc5d14ae7467d2e36e1ab4f0d8f9d059a3d071ce'
+    cpus: 1
+    mem_limit: '1g'
+    volumes:
+      - '../nginx:/etc/nginx'
+    ports:
+      - '0.0.0.0:7080:7080'
```
slimsag: Add a comment here describing how someone could alter this line to listen on port 80 instead.

docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,562 @@
...
+      - 'SRC_SYNTECT_SERVER=http://syntect-server:9238'
+      - 'SEARCHER_URL=http://searcher-0:3181'
+      - 'SYMBOLS_URL=http://symbols-0:3184'
+      - 'INDEXED_SEARCH_SERVERS=zoekt-webserver-0:6070'
+      - 'SRC_FRONTEND_INTERNAL=sourcegraph-frontend-internal:3090'
+      - 'REPO_UPDATER_URL=http://repo-updater:3182'
+      - 'REPLACER_URL=http://replacer:3185'
+      - 'LSIF_SERVER_URL=http://lsif-server:3186'
+      - 'GRAFANA_SERVER_URL=http://grafana:3370'
+    healthcheck:
```
slimsag: File an issue to move these into our Dockerfiles later

docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,562 @@
...
+      test: "wget -q 'http://127.0.0.1:9238/health' -O /dev/null || exit 1"
+      interval: 1s
+      timeout: 5s
+      retries: 1
+      start_period: 5s
+    networks:
+      - sourcegraph
+    restart: always
+  # Description: Backend for symbols operations.
```
slimsag:
```suggestion

    # Description: Backend for symbols operations.
```

slimsag:
I am content with this as-is for merging, please follow up on my two inline comments later.

I pushed up some basic docs which we can move to docs.sourcegraph.com later (these are temporary and only exist for purposes of showing this to some initial users who are interested in this deployment model). I am going to merge this now in order to show this to some users and get feedback.

PR #39, opened 2019-09-04
use custom prom/grafana images with shared provisioning

uwedeportivo:
images defined in https://github.com/sourcegraph/sourcegraph/pull/5343


deploy-grafana.sh
```diff
@@ -17,14 +20,14 @@ docker run --detach \
...
-    -v $(pwd)/grafana:/etc/grafana \
+    -v ${DIR}/grafana:/sg_config_grafana/provisioning/datasources \
     -e GF_AUTH_ANONYMOUS_ENABLED=true \
     -e GF_AUTH_ANONYMOUS_ORG_NAME=Sourcegraph \
     -e GF_AUTH_ANONYMOUS_ORG_ROLE=Editor \
     -e GF_USERS_ALLOW_SIGN_UP='false' \
     -e GF_USERS_AUTO_ASSIGN_ORG='true' \
     -e GF_USERS_AUTO_ASSIGN_ORG_ROLE=Editor \
-    grafana/grafana:6.1.1@sha256:e7a513bf7f33ef9681b2d35a799136e1ce9330f9055f75dfa2101d812946184b
+    sourcegraph/grafana:3.8
```
slimsag: Could you add the SHA for this (`docker inspect sourcegraph/grafana:3.8` will tell you it), to ensure the image doesn't change by accident if anyone e.g. force pushes the docker tag or something.

deploy-prometheus.sh
```diff
@@ -16,10 +20,6 @@ docker run --detach \
...
-    -v ~/sourcegraph-docker/prometheus-disk:/prometheus-data \
-    -v $(pwd)/prometheus:/etc/prometheus \
-    sourcegraph/prometheus:v1.4.1@sha256:1a02878ac9a0f532a68673bda93d406b21786706446c42a40cb3ed3e867278c0 \
-    -config.file=/etc/prometheus/prometheus.yml \
-    -storage.local.path=/prometheus-data \
-    -web.console.libraries=/etc/prometheus/console_libraries \
-    -web.console.templates=/etc/prometheus/consoles
+    -v ~/sourcegraph-docker/prometheus-disk:/prometheus \
+    -v ${DIR}/prometheus:/sg_prometheus_add_ons \
+    sourcegraph/prometheus:3.8
```
slimsag: same here re: sha hash

deploy-grafana.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
-set -e
+
+set -euf -o pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
slimsag: set -euf and the DIR change here should be undone. The goal of these scripts is to be super simple and make it clear how one would deploy all of these Sourcegraph containers in their own setup -- while these changes may be better in general for bash, in this case they distract I think from the intent here. In other words, `set -e` and `$(pwd)` is good enough in this repository.

prometheus/README.md
```diff
@@ -5,3 +5,5 @@ This directory contains configuration for the [Prometheus](https://prometheus.io
 This directory is mounted into the `prometheus` container. After making your changes to this directory,
 simply `docker restart prometheus` for your changes to take effect (depending on your change, Prometheus
 may respond to it as soon as you save the file).
+
+You can add `_rules.yml` and `_targets.yml` files to this directory and they will be picked up automatically.
```
slimsag: Can you document here exactly which file would need to be created in order to retain the Docker metric scraping behavior from before:

```
  # Scrape Docker for metrics.
  # You will need to configure docker for this to work: https://docs.docker.com/config/thirdparty/prometheus/#configure-docker
  - job_name: "docker"
    static_configs:
      # For Linux, use this line:
      #- targets: ["localhost:9323"]
      #
      # For Mac, use this line:
      #- targets: ["docker.for.mac.host.internal:9323"]
      #
      # For Windows, use this line:
      #- targets: ["docker.for.win.localhost:9090"]
```

deploy-grafana.sh
uwedeportivo: done

deploy-prometheus.sh
uwedeportivo: done

deploy-grafana.sh
uwedeportivo: done

prometheus/README.md
uwedeportivo: done

PR #40, opened 2019-09-11
Cleanup some monitoring changes

slimsag:
This PR cleans up some things around monitoring I observed from https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/39

- Handle upgrades so no manual intervention is needed (Prometheus v1 data is not compatible with v2).
- Add back the `docker.rules` which are needed for the "Containers" dashboard.
- Minor README/ whitespace cleanup.

Fixes sourcegraph/sourcegraph#5546

### DIFF
--- a/deploy-prometheus.sh
+++ b/deploy-prometheus.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-
 set -e
 
 # Description: Prometheus collects metrics and aggregates them into graphs.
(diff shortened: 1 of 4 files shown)


deploy-prometheus.sh
```diff
@@ -17,6 +16,6 @@ docker run --detach \
     --cpus=4 \
     --memory=8g \
     -p 0.0.0.0:9090:9090 \
-    -v ~/sourcegraph-docker/prometheus-disk:/prometheus \
+    -v ~/sourcegraph-docker/prometheus-v2-disk:/prometheus \
```
uwedeportivo: ah, yeah. good point. thanks.

PR #46, opened 2019-10-04
deploy automation

sqs:
WIP: This deploys the frontend with the automation private beta features.

It does not yet include the Ruby and Gradle services. We will be able to smoke test this deployment using automation campaigns that do not require these (such as the package.json dep removal and Comby campaigns), and we'll add those soon.


slimsag:
I had to base this on a different branch for deploying it -- don't worry about this but just bump the images in this branch instead: https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/47

PR #47, opened 2019-10-05
a8n test customer replica

slimsag:
@sqs This is a branch of `3.8-customer-replica` which you can push changes to for testing purposes. To deploy them to the replica:

```
ssh ec2-user@50.112.230.151
cd deploy-sourcegraph-docker/
./teardown.sh && git pull && ./deploy.sh
```


deploy-frontend.sh
```diff
@@ -31,7 +31,7 @@ docker run --detach \
     -e GRAFANA_SERVER_URL=http://grafana:3000 \
     -v ~/sourcegraph-docker/sourcegraph-frontend-$1-disk:/mnt/cache \
     -p 0.0.0.0:$((3080 + $1)):3080 \
-    sourcegraph/frontend:3.8.2
+    sourcegraph/frontend@sha256:a658bbe4cbb40ba16891d036bc2892c80c98a05c5c4184d336020e7ff6fab5cd
```
slimsag: note: I hard-coded these in place instead of using a separate file just to reduce the diff when shipping (to avoid any confusion/typos/etc)

PR #48, opened 2019-10-08
upgrade grafana

uwedeportivo:
- moves default port to 3370
- upgrades underlying grafana to 6.4.1
- fixes jsonnet dashboard generation for single server
- allows grafana to work outside the reverse proxy

related to https://github.com/sourcegraph/sourcegraph/pull/5910

### DIFF
--- a/deploy-grafana.sh
+++ b/deploy-grafana.sh
@@ -15,11 +15,12 @@ docker run --detach \
     --restart=always \
     --cpus=1 \
     --memory=1g \
-    -p 0.0.0.0:3000:3000 \
+    -p 0.0.0.0:3370:3370 \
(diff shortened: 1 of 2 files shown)


deploy-grafana.sh
```diff
@@ -15,11 +15,12 @@ docker run --detach \
     --restart=always \
     --cpus=1 \
     --memory=1g \
-    -p 0.0.0.0:3000:3000 \
+    -p 0.0.0.0:3370:3370 \
+    -e GF_SERVER_SERVE_FROM_SUB_PATH=true \
```
slimsag: Is there a reason this can't / shouldn't be the default value in our image?

uwedeportivo: My bad :-)

PR #51, opened 2019-11-15
monitoring: postgres metrics

uwedeportivo:
Exposes Prometheus target postgres_exporter that publishes Postgres metrics of an associated Postgres instance.

Related to https://github.com/sourcegraph/sourcegraph/pull/6616

### DIFF
--- a/deploy-postgres-exporter.sh
+++ b/deploy-postgres-exporter.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+# Description: Export Prometheus metrics from associated Postgres instance
+#
+# Disk: none
+# Network: 100mbps
+# Liveness probe: n/a
+# Ports exposed to other Sourcegraph services (Prometheus target): 9187/TCP
+# Ports exposed to the public internet: none
+#
+docker run --detach \
+    --name=postgres-exporter \
+    --network=sourcegraph \
+    --restart=always \
+    --cpus=1 \
+    --memory=1g \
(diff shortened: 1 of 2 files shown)


deploy-postgres-exporter.sh
```diff
@@ -0,0 +1,21 @@
...
+# Ports exposed to other Sourcegraph services (Prometheus target): 9187/TCP
+# Ports exposed to the public internet: none
+#
+docker run --detach \
+    --name=postgres-exporter \
+    --network=sourcegraph \
+    --restart=always \
+    --cpus=1 \
+    --memory=1g \
+    -e DATA_SOURCE_NAME="postgresql://sourcegraph:sourcegraphd@pgsql:5432/postgres?sslmode=disable" \
```
slimsag: @uwedeportivo Can a user specify PG* variables here individually instead of a DSN? Composing the DSN requires hard-coding the password or some weird trickery if it contains values that need to be URL-encoded

I imagine the same would be true in deploy-sourcegraph; we should probably just use the PG* variables mirroring the frontend

prometheus/prometheus_targets.yml
```diff
@@ -21,3 +21,5 @@
     - gitserver-0:6060
     - searcher-0:6060
     - symbols-0:6060
+    - postgres-exporter:9187
```
slimsag: This line should go below lsif-server on line 18; the area you put it are for entries that must be duplicated for multiple replicas

PR #62, opened 2020-01-15
docker-compose: helper script to update cpus

ggilmore:
I wrote this quick script to help me tune all of the `cpu` settings to something my machine can run. I don't know if this needs to be merged, but it's been really helpful to have this so that I can test everything out.


slimsag:
Probably fine to keep just as a PR for now, while we figure out what to do w.r.t. generating YAML etc.

PR #65, opened 2020-01-24
docker-compose: use docker volumes

keegancsmith:
Removes the need to bind mount to local directories. Potentially also removes
some permission errors encountered.

Fixes: https://github.com/sourcegraph/sourcegraph/issues/7829

Did this while experimenting for https://github.com/sourcegraph/sourcegraph/issues/7803


ggilmore:
@slimsag PTAL

This is essentially the same as my branch (except now that it works since I cut 3.12.3

PR #66, opened 2020-01-24
docker-compose: systemd service file

keegancsmith:
Part of https://github.com/sourcegraph/sourcegraph/issues/7803

### DIFF
--- a/docker-compose/sourcegraph.service
+++ b/docker-compose/sourcegraph.service
@@ -1,0 +1,29 @@
+### Installation
+#
+# 1. Replace "ec2-user" with your username.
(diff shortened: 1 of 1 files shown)


slimsag:
Given our docker-compose has `restart:always` in it (I think?) it seems we would not need this to persist across restarts, so why is this desirable / needed?

keegancsmith:
> Given our docker-compose has `restart:always` in it (I think?) it seems we would not need this to persist across restarts, so why is this desirable / needed?

This was requested by the customer. I wasn't aware restart:always had that behaviour. In that case I'd rather not support the systemd usecase officially, so lets just close this out.

PR #68, opened 2020-01-25
pgsql: mount volume to the correct data storage location in the c…

ggilmore:
The postgres volume should be mounted to `/data` instead of `/data/pgsql`: see https://github.com/sourcegraph/deploy-sourcegraph/blob/7e23bf8563b12353c7c920f21d595340e25efb65/base/pgsql/pgsql.Deployment.yaml#L51-L52 for how this is done in Kubernetes. 

Currently, it seems like **users would lose their data** if the postgres container was ever stopped and removed. I was able to verify this via [docker-compose](https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/65):

1. Run `docker-compose up -d` to start up a new instance.
1. Create an admin account 
1. Run `docker stop pgsql` and `docker rm pgsql` 
1. Run `docker-compose` restart - see that you aren't able to load the web-page (this is expected so far)
1. Run `docker-compose up -d` to create the pgsql container again and `docker-compose restart` to refresh everything 

Notice that this takes you back to the site initialization page. If the pgsql container really was writing everything properly to the volume, you would still be able to sign in. 

After I fixed the file paths in this PR and ran through these steps again, I was able to sign in at the end (what you expect when pgsql is properly saving its data to the volume).

### DIFF
--- a/deploy-pgsql.sh
+++ b/deploy-pgsql.sh
@@ -15,7 +15,7 @@ docker run --detach \
     --restart=always \
     --cpus=4 \
     --memory=2g \
-    -v ~/sourcegraph-docker/pgsql-disk:/data \
+    -v ~/sourcegraph-docker/pgsql-disk:/data/ \
     index.docker.io/sourcegraph/postgres-11.4:19-11-14_b084311b@sha256:072481559d559cfd9a53ad77c3688b5cf583117457fd452ae238a20405923297
 
 # Sourcegraph requires PostgreSQL 9.6+. Generally newer versions are better,
--- a/deploy-pgsql.sh
+++ b/deploy-pgsql.sh
@@ -15,7 +15,7 @@ docker run --detach \
     --restart=always \
     --cpus=4 \
     --memory=2g \
-    -v ~/sourcegraph-docker/pgsql-disk:/data/pgdata \
+    -v ~/sourcegraph-docker/pgsql-disk:/data \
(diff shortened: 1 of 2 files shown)


slimsag:
1. I don't know why it was set to `/data/pgdata` instead of `/data`, that looks like obviously a mistake.
2. I see our 9.4 and 9.6 images set PGDATA to `/data/pgsql`, which is why this worked in the past (I certainly tested this and didn't run into any data loss before): https://sourcegraph.sgdev.org/search?q=repo:%5Esourcegraph/infrastructure%24%40de225d0+/data/pgdata&patternType=literal
3. I didn't spot this earlier because the customer-replica branch I run, which mimics our largest customer using this, runs the 9.6 image to mirror their pgsql version: https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/3.10-customer-replica/deploy-pgsql.sh
4. Our 11.4 and 11.6 images appear to have changed the file path from `/data/pgdata` to `/data/pgdata-11` as part of our pgsql auto-migration script: https://sourcegraph.sgdev.org/sourcegraph/infrastructure/-/blob/docker-images/postgres-11.4/Dockerfile#L39-40

Thus, this regression was introduced in https://github.com/sourcegraph/deploy-sourcegraph-docker/commit/6e9b65308176fb5fafe193117d34de802bc523a7

Impact:

- All customers I am aware of that are using the deploy.sh scripts are currently using an external PostgreSQL instance, so they would not be affected.
- Recent adopters of the docker-compose deployment would be affected by this.

PR #82, opened 2020-03-13
Federate lsif-server metrics

efritz:
**Note**: This should not be merged until the docker tags have been updated in the deployment, otherwise we will lose metrics for LSIF processes as prometheus is not available in the current containers.

https://github.com/sourcegraph/sourcegraph/pull/8951 introduced a Prometheus server inside the lsif-server containers to scrape and federate metrics from a dynamic number of workers. This introduces a new annotation option `prometheus.io/federate` that enables the `/federate` endpoint to be scraped for all global metrics.


efritz:
Closing in favor of https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/93/files

PR #93, opened 2020-03-30
Rename lsif-server to precise-code-intel

efritz:
This is the pure docker/docker compose analog for https://github.com/sourcegraph/deploy-sourcegraph/pull/574.

Removes the lsif-server containers in favor of precise-code-intel-api-server, precise-code-intel-bundle-manager, and precise-code-intel-worker.


prometheus/prometheus_targets.yml
```diff
@@ -12,8 +12,8 @@
     - query-runner:6060
     - repo-updater:6060
     - syntect-server:6060
-    - lsif-server:3186
-    - lsif-server:3187
+    - precise-code-intel-api-server:3186
+    - precise-code-intel-bundle-manager:3187
```
slimsag, line 16: Where is the entry or the worker?

docker-compose/docker-compose.yaml
```diff
@@ -252,29 +246,76 @@ services:
...
+  # Ports exposed to other Sourcegraph services: 3188/TCP (first worker), 9090 (prometheus)
+  # Ports exposed to the public internet: none
+  #
+  precise-code-intel-worker:
+    container_name: precise-code-intel-worker
+    image: 'index.docker.io/sourcegraph/precise-code-intel-worker:insiders'
+    cpus: 2
+    mem_limit: '4g'
+    environment:
+      - NUM_WORKERS=4
```
slimsag, line 310: Can we rely on the default (2) value here instead?

prometheus/prometheus_targets.yml
efritz: It's federated in prometheus as it can run more than one metrics endpoint.

docker-compose/docker-compose.yaml
efritz: No problem.

docker-compose/docker-compose.yaml
```diff
@@ -252,29 +246,76 @@ services:
...
+  #
+  precise-code-intel-bundle-manager:
+    container_name: precise-code-intel-bundle-manager
+    image: 'index.docker.io/sourcegraph/precise-code-intel-bundle-manager:insiders'
+    cpus: 2
+    mem_limit: '2g'
+    environment:
+      - PRECISE_CODE_INTEL_API_SERVER_URL=http://precise-code-intel-api-server:3186
+      - 'SRC_FRONTEND_INTERNAL=sourcegraph-frontend-internal:3090'
+      - LSIF_STORAGE_ROOT=/lsif-storage
```
slimsag, line 286: Why is this here?

deploy-precise-code-intel-bundle-manager.sh
```diff
@@ -0,0 +1,23 @@
...
+#
+docker run --detach \
+    --name=precise-code-intel-bundle-manager \
+    --network=sourcegraph \
+    --restart=always \
+    --cpus=2 \
+    --memory=2g \
+    -e 'PRECISE_CODE_INTEL_API_SERVER_URL=http://precise-code-intel-api-server:3186' \
+    -e 'SRC_FRONTEND_INTERNAL=sourcegraph-frontend-internal:3090' \
+    -e 'LSIF_STORAGE_ROOT=/lsif-storage' \
```
slimsag, line 19: shouldn't be here

deploy-precise-code-intel-worker.sh
```diff
@@ -0,0 +1,21 @@
...
+# Ports exposed to other Sourcegraph services: 3188/TCP (first worker), 9090 (prometheus)
+# Ports exposed to the public internet: none
+#
+docker run --detach \
+    --name=precise-code-intel-worker \
+    --network=sourcegraph \
+    --restart=always \
+    --cpus=2 \
+    --memory=4g \
+    -e 'NUM_WORKERS=2' \
```
slimsag, line 16: shouldn't be here, we should be able to rely on the default being 2

docker-compose/docker-compose.yaml
efritz: The bundle manager still has a volume mounted at /lsif-storage. We can remove this as it's the default value for that envvar, but 294 would need to stay the same.

deploy-precise-code-intel-worker.sh
efritz: Do you mean the default in the docker container?

deploy-precise-code-intel-bundle-manager.sh
efritz: See above.

deploy-precise-code-intel-worker.sh
slimsag: Yes

efritz: I'm removing all NUM_WORKERS from this repo and will change the image default to 2 in sg/sg.

docker-compose/docker-compose.yaml
slimsag: Makes sense, yes we should rely on the default value.

deploy-precise-code-intel-bundle-manager.sh
slimsag: We should rely on the default value.

docker-compose/docker-compose.yaml
```diff
@@ -252,29 +246,75 @@ services:
...
+
+  # Description: Stores and manages precise code intelligence bundles.
+  #
+  # Disk: 200GB / persistent SSD
+  # Ports exposed to other Sourcegraph services: 3187/TCP (manager)
+  # Ports exposed to the public internet: none
+  #
+  precise-code-intel-bundle-manager:
+    container_name: precise-code-intel-bundle-manager
+    image: 'index.docker.io/sourcegraph/precise-code-intel-bundle-manager:insiders'
```
slimsag, line 280: It's OK to merge this with `insiders` tags, but they need to be pinned using SHA digests.

docker-compose/docker-compose.yaml
```diff
@@ -252,29 +246,75 @@ services:
...
+  # Ports exposed to other Sourcegraph services: 3188/TCP (first worker), 9090 (prometheus)
+  # Ports exposed to the public internet: none
+  #
+  precise-code-intel-worker:
+    container_name: precise-code-intel-worker
+    image: 'index.docker.io/sourcegraph/precise-code-intel-worker:insiders'
+    cpus: 2
+    mem_limit: '4g'
+    environment:
+      - PRECISE_CODE_INTEL_BUNDLE_MANAGER_URL=http://precise-code-intel-bundle-manager:3187
```
slimsag, line 310: For consistency, the env vars you added to this file should also be in single quotes like all others in this file

efritz: Done.

docker-compose/docker-compose.yaml
efritz: Updated to same commit as the others.

deploy-precise-code-intel-bundle-manager.sh
efritz: Removed.

prometheus/prometheus.yml
```diff
@@ -0,0 +1,47 @@
...
+  - job_name: "sg"
+    file_sd_configs:
+      - files:
+          - "/sg_prometheus_add_ons/*_targets.yml"
+
+## Note: the content above is taken from sourcegraph/sourcegraph/docker-images/prometheus/config/prometheus.yml
+## I'm not currently sure how to configure the below in an external file via file_sd_configs.
+##
+## TODO(eric,uwedeportivo) - find a way to express what's below in an extensible way so we can 
+## just add config files in this directory instead of overwriting the entire main configuration.
```
slimsag, line 36: We can't merge this PR as-is with this in place. We must move towards less configuration in this repository and not more, introducing a new config file here is IMO not an option.

A few options I would be content with:

1. Merge this PR without monitoring for the workers, and defer this work to another PR. I am fine with that, but we would need to figure it out before 3.15 releases.
2. An environment variable on the prometheus image which controls whether or not this configuration is loaded (not a file stored in this repo).

efritz: We can remove monitoring, but I'm not 100% we can fix this before the 3.15 release in the way @sourcegraph/code-intel wants (we're trying to rewrite parts of the worker in Go so we don't have this problem which is caused by running multiple node processes in the same container).

cc @uwedeportivo

slimsag: I don't think it would be too hard to solve, though?

Right now in `master` we are using the builtin `prometheus.yml` here: https://github.com/sourcegraph/sourcegraph/blob/master/docker-images/prometheus/config/prometheus.yml

You would just need to copy this prometheus.yml into the same location, but name it say e.g. `prometheus_pure_docker.yml` and then use it by setting the env var here: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@6fef3d27e4938dd8d750afd7a81dd6a150d04e9b/-/blob/docker-images/prometheus/entry.sh#L4-8

That wouldn't be perfect (because you would still have `precise-code-intel-worker:9090` hard-coded in that file) but that'd be good enough for 3.15.

Then in e.g. `deploy-prometheus.sh` in this repo you would just add `-e PURE_DOCKER=true` to trigger using that config file.

efritz: Ok that seems like a fine stopgap.

efritz: Working on this now. And for context, this entire federation technique should be able to be removed in a few cycles once we rewrite some of the worker wrapper in Go and can use the language concurrency rather than the OS's.

efritz: See https://github.com/sourcegraph/sourcegraph/pull/9552.

I'm unsure of how to update the prometheus image, though.

PR #94, opened 2020-04-13
docker-compose: add jaeger-all-in-one

beyang:
Adds the jaeger-all-in-one container to the pure Docker and Docker Compose configuration. Also fixes a few broken things in the pure Docker scripts:
- Invalid comment syntax broke `deploy-frontend.sh`
- Permissions issues with any service that bind-mounts a volume from the host

TODO
- [x] Address https://github.com/sourcegraph/deploy-sourcegraph/pull/559#discussion_r405182789

Post-merge TODO
- [ ] Address https://github.com/sourcegraph/deploy-sourcegraph/pull/559#discussion_r405181298


slimsag:
It's not OK for Docker Compose and Pure Docker (the shell scripts at the root of this repository) to get more out of sync. Please update those to match this change (should be easy / mostly copy+paste)

slimsag:
LGTM aside from that

deploy-frontend-internal.sh
```diff
@@ -26,9 +26,9 @@ docker run --detach \
     -e SRC_FRONTEND_INTERNAL=sourcegraph-frontend-internal:3090 \
     -e REPO_UPDATER_URL=http://repo-updater:3182 \
     -e REPLACER_URL=http://replacer:3185 \
-    # remove after 3.15
+    `# remove after 3.15` \
```
slimsag, line 29: I am surprised this doesn't cause the `docker` command to break or complain about an extra argument?

beyang: On the contrary, this is required to fix the very issue you're pointing out. Bash doesn't permit comments in the midst of multi-line constructions, and the backtick syntax is a way to "trick" it into interpreting this as just another line.

PR #105, opened 2020-05-13
Increase searcher max memory usage

rvantonder:
This is a proposed change, see https://sourcegraph.slack.com/archives/CJX299FGE/p1588960025377500 for more context.

2g is quite low for searcher on small/medium deployments, in the context of structural search/comby for large repos. Although comby does not itself perform super memory-intensive operations, it will load a repo zip file content into memory while searching, and 6g is a better mem limit. It will otherwise OOM.

Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change:

There is no docker-compose/... folder in deploy-sourcegraph repo. I don't know if something is needed there.

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -209,7 +209,7 @@ services:
     container_name: searcher-0
     image: 'index.docker.io/sourcegraph/searcher:3.15.1@sha256:de5601ae6446c37b7f49adf3de8de25475c700a77f31dee8d28139fef0dc6584'
     cpus: 2
-    mem_limit: '2g'
+    mem_limit: '6g'
     environment:
       - GOMAXPROCS=2
       - 'SRC_FRONTEND_INTERNAL=sourcegraph-frontend-internal:3090'


slimsag:
Thanks for re-flagging this, apologies I forgot to respond to it!

Your intuition that they should increase the memory limit is definitely correct (especially if it is OOMing), however, I think most deployments these days are not using Comby search and when they are it is often not with very large repositories.

If we can get rough confidence that >50% of instances do need more memory for searcher, we can definitely increase the base requirements of it. Right now we do not have a good way to gather this information -- but hoping to have something like this in the coming months where we get ping data about customer's CPU/memory usage.

For now, I think we are better off suggesting customers increase the memory of this service ad-hoc when they run into problems.

> Sister deploy-sourcegraph change:
>
> There is no docker-compose/... folder in deploy-sourcegraph repo. I don't know if something is needed there.

Ah, fixed -- this was supposed to refer to "updating the kubernetes yaml"

rvantonder:
> If we can get rough confidence that >50% of instances do need more memory for searcher, we can definitely increase the base requirements of it. 

I'm looking more at existing services and thumb sucking relative to that. If we are allocating up to 8g just for precise-intel-* workers (which may also not be active in deployments to day, or empirically validated), then 2g for a front-facing feature we're putting at the fingertips of a user in the UI feels low.

The other way I'm thinking about this is: if I rearchitected searcher to temporarily unzip repo contents to disk, comby would happily run with 2g memory by processing things on disk. If there are more incoming reports about OOM I'll probably consider doing that, because even if it's slower, it sure beats "feature unavailable". I'm not sure that admins will be responsive about ad-hoc increasing memory in general, and it's unfortunate that a user has to hit that first for it to happen. In this case I'd prefer a a strategy that doesn't increase friction for any customer (not just a majority) to bolster the chances of usage for an admittedly new/unfamiliar feature, at the cost of perf/memory use.

slimsag:
Now we get to the fun stuff :)

So, our repositories aren't really in sync. We have three sets of resource recommendations:

1. The "bare minimum" recommendations
 in https://github.com/sourcegraph/deploy-sourcegraph/tree/master/base which reserves just 1.5G for all precise-code-intel services (500M each).
2. The "production instance" recommendations in this repository which are heavily based on a certain $CUSTOMER who runs 18 searcher instances currently, noting that the values in this repository are limits _not allocations_.
3. The resource recommendations in the estimator https://docs.sourcegraph.com/admin/install/resource_estimator which tries to supersede both of these by giving estimates that are based on actual known-good-customer-deployments and scaling accordingly.

Now, none of these three have a good way to express resources that might not get used. For example, as you noted, most customers today simply aren't going to use precise-code-intel -- so it's not really clear if we should recommend almost _no resources_ for that deployment out of the box, or _some_, or _the max we've ever seen it require with heavy usage_. We just don't have a good way of expressing this today, aside from "If it OOMs, you should be giving it more memory" as a rule of thumb.

> If we are allocating up to 8g just for precise-intel-* workers (which may also not be active in deployments to day, or empirically validated), then 2g for a front-facing feature we're putting at the fingertips of a user in the UI feels low.

I can see how you arrived here, but these limits were actually just blindly copied from https://github.com/sourcegraph/deploy-sourcegraph/tree/master/base/precise-code-intel knowing that since it's a new feature it will see little to no usage and thus poses no risk to existing deployments.

> The other way I'm thinking about this is: if I rearchitected searcher to temporarily unzip repo contents to disk, comby would happily run with 2g memory by processing things on disk. If there are more incoming reports about OOM I'll probably consider doing that, because even if it's slower, it sure beats "feature unavailable".

It is 100% valid for searcher to use 6GB of memory (or even 8G of memory), if that is what makes sense in the code today. What I mean with this is that we certainly do not need to rearchitect searcher to avoid asking admins to give us ~4-6GB more of memory.

What we need is the following:

- Our baseline resource guidance (in these repos, and the resource estimator) give you what 90% of our customers use. This means, that things which are not used frequently _today_ by 90% of our users (precise code intel and Comby search fall into this category today) will not strictly have all the resources they need out-of-the-box to run at maximum capacity -- but that's okay because for 90% of our customers they don't use it at maximum capacity.
- Our alerting and documentation give you guidance on what to do when that is not the case, when your users start using precise-code-intel (at all, or heavily) or comby search. We should feel very comfortable saying "yes, it's OOMing so you need to give it more memory" unless we're asking for something crazy.

> I'm not sure that admins will be responsive about ad-hoc increasing memory in general, and it's unfortunate that a user has to hit that first for it to happen. In this case I'd prefer a a strategy that doesn't increase friction for any customer (not just a majority) to bolster the chances of usage for an admittedly new/unfamiliar feature, at the cost of perf/memory use.

Admins can and will be responsive to increasing resources when their Sourcegraph instance needs them. We just have to tell them this is the case. One part of this is surfacing when services are running out of resources to admins, which is what https://github.com/sourcegraph/sourcegraph/issues/9791 aims to solve, and another part of this is surfacing "structural searches are failing" to admins: we should add a Prometheus metric for exactly this and have it in the frontend dashboard of Grafana.

I guarantee you this: admins that are paying for Sourcegraph, when told something is wrong with Sourcegraph, will absolutely want it fixed.

At a higher level, what I am trying to say is: How much memory does searcher need to run well in _most_ deployments? Right now, we have no empirical evidence for this (as you noted) and it's hard to determine this. The best we can use for now is (1) the current status quo and whether or not customers complain (N=1 so far) and (2) what little data we can gather from sourcegraph.com, k8s.sgdev.org, and the few customer instances we have access to.

It's definitely not your fault we're in this mess (it's actually mostly mine) -> but I promise the way to fix this is by (1) telling admins to give it more memory until we have more evidence most deployments need it and (2) putting in place alerts which make it super clear "Comby search is failing, you should fix this" and we _do_ have the infrastructure in place to do this :)

rvantonder:
OK thanks for the detailed response. I'm going to take care by raising the right level of awareness in alerts for this in the app.

PR #116, opened 2020-06-02
Add env var to frontend deployment for Jaeger

daxmc99:
Fixes https://github.com/sourcegraph/deploy-sourcegraph/pull/716

Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change:https://github.com/sourcegraph/deploy-sourcegraph/pull/716


deploy-pgsql.sh
```diff
@@ -15,7 +15,6 @@ docker run --detach \
     --name=pgsql \
     --network=sourcegraph \
     --restart=always \
-    --cpus=4 \
```
bobheadxi, line 18: why was this removed? :o

daxmc99: I was running the deploy scripts locally and kept seeing some CPU issues, this should not be here

PR #117, opened 2020-06-02
Add caddy to teardown script

daxmc99:
Caddy was not being removed on teardown

### DIFF
--- a/deploy-caddy.sh
+++ b/deploy-caddy.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+set -e
+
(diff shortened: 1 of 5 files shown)


slimsag:
There actually isn't a `deploy-caddy.sh` script so I am guessing you had it left over from something else (maybe the docker-compose deployment).

It would be great to add this and a `deploy-caddy.sh` script, though, so we can keep docker-compose and pure-docker in sync (they should be nearly identical)

daxmc99:
Yeah, I'm now realizing that. 
I will add a deploy caddy script

PR #121, opened 2020-07-01
WIP: Split docker-compose code into multiple files

pecigonzalo:
This is a WIP PR as an example of how to split our compose files into multiple sections.

Usage:
```
# AIO
docker-compose -p FOO \
  -f docker-compose/docker-compose.frontend.yaml \
  -f docker-compose/docker-compose.db.yaml \
  -f docker-compose/docker-compose.core.yaml \
  -f docker-compose/docker-compose.monitoring.yaml \
up

# No monitoring
docker-compose -p FOO \
  -f docker-compose/docker-compose.frontend.yaml \
  -f docker-compose/docker-compose.db.yaml \
  -f docker-compose/docker-compose.core.yaml \
up
```

cc/ @sourcegraph/distribution


bobheadxi:
Unfortunately this approach seems like it would remove the possibility of a one-liner, ie 

```
curl docker-compose.yml | docker-compose -f -
```

[`-f -` to read from stdin](https://docs.docker.com/compose/reference/overview/#use--f-to-specify-name-and-path-of-one-or-more-compose-files)

pecigonzalo:
If we need a one liner we can do `curl this | sh -c` to a script.

PR #122, opened 2020-07-02
add SRC_FRONTEND_INTERNAL to prometheus, remove from grafana

bobheadxi:
follow-up to https://github.com/sourcegraph/sourcegraph/pull/11832

Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: n/a, frontend is already configured to find prometheus and vice versa


slimsag:
CI is failing due to prettier on the Markdown files which we don't want to enforce, you can add the Markdown files in this repo to `.prettierignore` so CI gets fixed. My bad, sorry!

PR #127, opened 2020-08-07
pure docker / customer replica smoke tests

slimsag:
### What this is

This PR adds basic smoke tests for releasing the pure-docker deployment (not docker-compose), which helps in shipping Sourcegraph on-time to the only customer using this deployment mode, https://app.hubspot.com/contacts/2762526/company/407948923/

Effectively, this performs the same very basic tests I have performed in all ~4-5 prior releases to them:

- Have any services' file permissions / static UID/GIDs changed?
- Do the Docker containers all come up OK?
- Do the Docker containers run for a few minutes OK?
- Is the frontend reachable?

### What this is not

One aspect this does NOT test is:

- Does upgrading from the prior version to the current version work?

It seems OK not to test this because: (1) file permission checking is done explicitly here (which has always been the largest risk with upgrades for the customer in question) and (2) more general upgrade issues should already be caught as part of our release process for other deployment methods, i.e. there is nothing unique to test in this specific environment and we can rely on the testing we do for k8s/server instead.

### Impacts this has in the longer term

This reiterates/places more emphasis and burden on us as an engineering organization (outside Distribution) to do more proper testing, however. It means I am no longer assessing per-release if additional testing is needed by myself before shipping to https://app.hubspot.com/contacts/2762526/company/407948923/ - in practice this hasn't been required in the last 5-6 months so I feel OK doing this.

### Impacts this has on the release process

This places slightly more burden on the person releasing Sourcegraph, removing that burden from me and enabling us to ship Sourcegraph on-time on the 20th to the customer in question (instead of a few weeks later, as has historically happened.)

This does not reduce the testing required for Docker Compose releases, i.e. [the manual smoke tests in our release process](https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/RELEASING.md#3-smoke-test-ensure-docker-compose-starts-from-scratch) today. However, this can be duplicated to do that in a more automated fashion at a later date.

### Not yet on CI

Note: This runs in ~10m on my 2019 Macbook Pro, but on CI takes a very long time and fails often due to some network issues. For now, it is disabled on CI. See https://github.com/sourcegraph/sourcegraph/issues/12996


.buildkite/hooks/pre-command
```diff
@@ -3,5 +3,11 @@
 set -eu
 cd $BUILDKITE_BUILD_CHECKOUT_PATH
 
+if ! command -v asdf &> /dev/null
+then
+    echo "skipping asdf, TODO: ask Dave to add asdf to baremetal image"
```
slimsag, line 8: @davejrt can we pre-install the asdf command in the baremetal agent image? I suspect we'll need this in most CI pipelines that use it, so it would make sense to bake it in

RELEASING.md
```diff
@@ -77,29 +77,22 @@ docker ps
 
 Visit http://localhost and confirm the app loads.
```
slimsag, line 78: These instructions not yet updated to reflect using this in CI.

.buildkite/hooks/pre-command
slimsag: Going to leave this as-is until https://github.com/sourcegraph/sourcegraph/issues/12996 is fixed (which is higher priority).

slimsag:
Merging without review so I can move onto the next part of this, which is https://github.com/sourcegraph/customer/issues/73 - I will follow up on all feedback left here.

PR #132, opened 2020-08-14
Sg/dave vagrant

davejrt:


slimsag:
@davejrt I merged my PR into master just now (with some other fixes), you can test by just uncommenting the following: https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/.buildkite/pipeline.yaml#L7-L14

PR #141, opened 2020-08-21
e2e testing

davejrt:
This PR is an attempt at the following:

Merge the smoke tests previously written by @slimsag

Enable testing on the baremetal agents using vagrant for a clean build environment

Some things to note:

There are currently two vagrant providers configured (vagrant-google and virtualbox). Virtualbox runs the vagrant box locally on the baremetal agent, and can be used to run tests locally as well as in CI. vagrant-google will provision a GCE instance and run the tests on that server, in the event that performance is an issue or a local dev environment does not have sufficient CPU and memory.

This PR in draft uses 8cpu and 8GB of RAM for the virtualbox. After testing to address sourcegraph/sourcegraph#12996 the best I can determine is that there's an issue with virtualbox in this environment and using anymore than 8vcpus seems to degrade performance to the point where these tests don't run. As a result as part of the test a maximum of 8 cpus are used for the zoekt services in order to actually run the tests. With no more than 8vcpus I have been able to reliably have tests run in ~16 minutes and 10 minutes using the vagrant-google plugin (which runs fine without the need to modify any cpu counts on services).


test/pure-docker/servers.yaml
```diff
@@ -0,0 +1,27 @@
+---
+-
+  box: google/gce
+  machine_type: "custom-16-20480"
+  name: pure-docker-test-gcp
+  shell_commands:
```
pecigonzalo, line 6: Ideally, we should make the scripts be the same for both except maybe for some "init"

davejrt: is this in reference to the `sed` command when using the virtualbox solution? I agree there is a lot of repeated code here. We can probably run at the majority of this as a shell block in the `Vagrantfile` then any other commands from the yaml afterwards which would be customised to google or vbox specifically?

pecigonzalo: It was mostly due to the repeated code and to avoid the chance of us forgetting to update one of the sections.

I like your either, we have the generic commands on the Vagrant file (or here in another section) and the box specific pre-commands as they are right now.

.buildkite/pipeline.yaml
```diff
@@ -2,13 +2,19 @@ $schema: 'http://json.schemastore.org/buildkite'
 steps:
   - label: ':docker:'
     command: .buildkite/validate-docker-compose.sh
+    agents:
+      queue: 'baremetal'
+    env:
+      PATH: "/var/lib/buildkite-agent/.asdf/shims:/var/lib/buildkite-agent/.asdf/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"
```
slimsag, line 8: Can't we modify the path in the agent image instead? Seems odd to do this here

deploy-precise-code-intel-bundle-manager.sh
```diff
@@ -8,6 +8,8 @@ source ./replicas.sh
 # Ports exposed to other Sourcegraph services: 3187/TCP
 # Ports exposed to the public internet: none
 #
+VOLUME="$HOME/sourcegraph-docker/lsif-server-disk"
+./ensure-volume.sh $VOLUME 999
```
slimsag, line 12: Why has this changed?

.tool-versions
```diff
@@ -1 +1,2 @@
 yarn 1.19.2
+nodejs 13.12.0
```
slimsag, line 2: Why has this changed?

slimsag:
I would like to see if we can get to the bottom of https://github.com/sourcegraph/sourcegraph/issues/12996 before merging this.. it worries me that we can't figure out why virtualbox is broken there and that we would have two different execution modes for these tests. I am worried this means the tests will end up not being reliable and we'll chalk up failures to infrastructure issues without understanding why in the same way we have done with e.g. DIND in the past.

slimsag:
> the best I can determine is that there's an issue with virtualbox in this environment and using anymore than 8vcpus seems to degrade performance to the point where these tests don't run.

https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/135 has one possible explanation for why this might have been the case. Is it possible you tested before or after this was merged? In that PR I found that concurrent `docker pull` requests would lead to a deadlock in Docker that caused the tests to fail, I believe we were hitting this on CI and locally (🤦)

.buildkite/pipeline.yaml
davejrt: the agent would not pick up the PATH defined on the physical server. It is defined as part of the build image, but even when checking the path as part of the PATH is not picked up.

slimsag: Oh interesting, I see

.tool-versions
davejrt: without this the prettier-check fails

deploy-precise-code-intel-bundle-manager.sh
davejrt: without this, by default I am getting the following errors during the tests:
```
    pure-docker-test-local: TEST FAILURE: expected 23 containers running, found 22
```

```
root@vagrant:~# docker logs -f f35c284a2fef
2020/08/24 23:36:10 failed to prepare directories: mkdir /lsif-storage/uploads: permission denied
```

.buildkite/pipeline.yaml
davejrt: note, this is all running as buildkite-agent at present, which I figured was a better approach than running as root.

davejrt:
> > the best I can determine is that there's an issue with virtualbox in this environment and using anymore than 8vcpus seems to degrade performance to the point where these tests don't run.
> 
> #135 has one possible explanation for why this might have been the case. Is it possible you tested before or after this was merged? In that PR I found that concurrent `docker pull` requests would lead to a deadlock in Docker that caused the tests to fail, I believe we were hitting this on CI and locally (🤦)

I previously rebased and made sure these changes were in these tests. I've confirmed that back on the 16cpu/20gb vagrant box it's still painfully slow. The [last test](https://buildkite.com/sourcegraph/deploy-sourcegraph-docker/builds/1002#3045df79-9ad5-4318-b575-8249270eff39) run with full specs ran for 45 minutes and was stalled for a while before I stopped it.

deploy-precise-code-intel-bundle-manager.sh
davejrt: I"m not sure why this doesn't have a the volume assigned to a user, given every other service we have that uses a volume runs this script. I've changed to the sourcegraph user as per the other services that run under that account. 

Related PR [here](https://github.com/sourcegraph/sourcegraph/pull/13383) should this get merged.

pecigonzalo:
FYI, something is running the tests twice, I couldnt find where on a quick look.

- https://buildkite.com/sourcegraph/deploy-sourcegraph-docker/builds/1011#4172feda-9ade-4f90-a23c-36f9ceefb56e/23-938
- https://buildkite.com/sourcegraph/deploy-sourcegraph-docker/builds/1011#4172feda-9ade-4f90-a23c-36f9ceefb56e/23-1142

davejrt:
> FYI, something is running the tests twice, I couldnt find where on a quick look.
> 
> * https://buildkite.com/sourcegraph/deploy-sourcegraph-docker/builds/1011#4172feda-9ade-4f90-a23c-36f9ceefb56e/23-938
> * https://buildkite.com/sourcegraph/deploy-sourcegraph-docker/builds/1011#4172feda-9ade-4f90-a23c-36f9ceefb56e/23-1142

@pecigonzalo Should be fixed now. I got rubocopped and stupidly put the config block inside the srv which means it got read twice. An error I swear I've made a thousand times in the past and still bites me sometimes.

davejrt:
So a bit of a re-think after meeting with @pecigonzalo this morning. Just one box now, set up to run locally unless the `VAGRANT_RUN_ENV="CI"` is set so local dev's just need to pass in their GCP username and they can test in our `sourcegraph-dev` project. Otherwise this now runs in a new `sourcegraph-ci` project which is separate from our regular dev work. 

Hopefully this workflow makes life a bit easier for devs and they don't have to worry about editing files and changing usernames/networks

pecigonzalo:
Just taking a look at the numbers, the test is taking 10 minutes, of which 5 minutes are "validating" over a period of 5 minutes. I think its quite nice time wise for an e2e.

.buildkite/test-pure-docker.sh
```diff
@@ -4,6 +4,12 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
 set -euxo pipefail
 
 cd test/pure-docker/
-vagrant destroy -f pure-docker-test
-vagrant up
+
+if ! vagrant plugin list --no-tty | grep vagrant-google; then
+
```
slimsag, line 9:
```suggestion
```

.buildkite/test-pure-docker.sh
```diff
@@ -4,6 +4,12 @@ cd "$(dirname "${BASH_SOURCE[0]}")/.."
...
+	vagrant plugin install vagrant-google
+
```
slimsag, line 11:
```suggestion
```

README.md
```diff
@@ -83,6 +83,48 @@ Then try cloning the repository:
 $ git clone ssh://git@myhost.com/my/repo /tmp/my-repo
 ```
 
+## Testing
```
slimsag, line 86: Consumers of this repository (people trying to deploy on Docker Compose, or pure docker) won't care about this so it shouldn't be in the README but rather in RELEASE.md. I will move.

test/pure-docker/Vagrantfile
```diff
@@ -1,68 +1,60 @@
...
+        o.ssh.username = username
+        o.ssh.private_key_path = ssh_key_path
+      end
+
+      srv.vm.provision 'shell', inline: <<-SHELL
+        #!/usr/bin/env bash
+        set -euxo pipefail
+        apt-get update
+        # Install Docker
+        curl -fsSL get.docker.com | sudo sh
```
slimsag, line 52: We need to install Docker using the method I had in use here originally below, I am worried there will end up being a difference between what this script does and what is in the apt packages

test/pure-docker/Vagrantfile
```diff
@@ -1,68 +1,60 @@
 # -*- mode: ruby -*-
```
slimsag, line 1: I was planning to copy this entire file before to create a new `test/docker-compose/Vagrantfile` test, is that still a sane approach given that this is now more templating logic than anything else? (everything is done in `servers.yaml` now)

test/pure-docker/servers.yaml
```diff
@@ -0,0 +1,15 @@
...
+  external_ip: false
+  use_private_ip: true
+  network: default
+  username: buildkite
+  ssh_key_path: "~/.ssh/id_rsa"
+  shell_commands:
+   - |
+      cd /deploy-sourcegraph-docker
+      ./deploy.sh
+   - "/deploy-sourcegraph-docker/test/pure-docker/smoke-test.sh"
```
slimsag, line 15: It looks like you have removed all of the logic I had for ensuring that permissions on volume directories hasn't changed: https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/141/files#diff-b584b34513d743c4570b69a661750e14L37-L64

We need to add this back, we can't remove it because this is a really important property to test for. The customer using this, https://app.hubspot.com/contacts/2762526/company/407948923/ does not have the `./ensure-volume.sh $VOLUME 100` invocations and that must be communicated through the upgrading docs as I said here: https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/141/files#diff-b584b34513d743c4570b69a661750e14L50-L51

slimsag:
@davejrt @pecigonzalo I am not sure I follow what the motivation here is for using different GCP projects when running this locally vs. in CI - what is the rationale for that? I think it would make more sense to have it always run in the same GCP project always

README.md
```diff
@@ -83,6 +83,48 @@ Then try cloning the repository:
...
+```
+
+- Configure your local variables using the following environment variables
+  - `VAGRANT_GCP_PROJECT_ID`: Project to run on. (default: `sourcegraph-server`)
+  - `VAGRANT_SSH_USER`: Your SSH user ID as specified in GCP metadata. (default: `ENV['USER']`)
+  - `VAGRANT_SSH_KEY`: Path to your SSH Keys as specified in GCP metadata. (default: `~/.ssh/id_rsa`)
+
+- Run the tests
+```
+VAGRANT_SSH_USER=ssh_user vagrant up pure-docker-test --provider=google
```
slimsag, line 107: This needs to mention cd'ing into the right directory first

README.md
slimsag: I think the best thing to do would be:

1. Move these instructions to a new `TESTING.md` file.
2. Update these instructions to refer to invoking `.buildkite/test-pure-docker.sh` instead of invoking `vagrant` directly (then you do not need to update `RELEASING.md` with what is effectively the same set of commands to run vagrant up and down.)

slimsag:
Looks pretty good, some things I'd like you to follow up on and then I'll re-review

pecigonzalo:
> @davejrt @pecigonzalo I am not sure I follow what the motivation here is for using different GCP projects when running this locally vs. in CI - what is the rationale for that? I think it would make more sense to have it always run in the same GCP project always

The main intention was to remove it from the production project, and then a second intention was having them in `-auxiliary` or something else than `-ci` would make it easier to distinguish dev created test vagrant machines from machine ones. 
I assume the concern about using a different project is the environment drift for user tests. At the moment I dont think it will cause any issues given the test is self contained, but is possible. We could also use a tag/metadata for that last bit and share `-ci`.

test/pure-docker/servers.yaml
davejrt: I have a PR with the information pending [here](https://github.com/sourcegraph/sourcegraph/pull/13383)

All of the deploy-<service>.sh scripts that require a volume now envoke `ensure-volume.sh`. Is this customer running a specific fork that doesn't include that? If so, wouldn't it make more sense for them to have that script?

test/pure-docker/Vagrantfile
davejrt: I would suggest that the Vagrant file remain as is, or install everything we need for both, then any excecution be done via the shell commands in the `servers.yaml`.

README.md
davejrt: Full instructions on setting up a local test are now in `TESTING.md`

test/pure-docker/Vagrantfile
davejrt: updated in Vagrant file

test/pure-docker/servers.yaml
slimsag: I am not sure how to avoid this confusion, but see the README of this repository for an explanation of what pure-docker deployments actually are:

> The goal is that anyone using a container management platform other than Kubernetes (Netflix's Titus, Apache's Mesos, etc.) would be able to use this repository as a **reference** for how to deploy Sourcegraph.

The point being the customer in question does NOT run the scripts in this repository, and any time there are changes in volume permissions etc. they need to know about those precisely because it means they need to be able to adapt those to their deployment in their own way. It's not a fork - it's the way they were intended to consume this repository (as a reference)

slimsag: The point being: we need [these lines](https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/141/files#diff-b584b34513d743c4570b69a661750e14L37-L65) to still exist in this test somewhere, removing them means we can't in good faith release pure docker to this customer and the test becomes effectively useless

RELEASING.md
```diff
@@ -79,9 +79,9 @@ Visit http://localhost and confirm the app loads.
 
 ## 5) Confirm Pure-Docker works with a smoke test
 
-This will take about ~10 minutes to run, and you must have a machine with 16 CPUs.
+Refer to the [TESTING.md](TESTING.md) for full instructions on running local tests.
```
slimsag, line 82:
```suggestion
```

slimsag:
Separation of dev and CI assets for cost-accounting purposes makes sense, I didn't think of that. Thanks!

slimsag:
Last thing to address here is https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/141#discussion_r479578063 I think

test/pure-docker/servers.yaml
slimsag: The most obvious solution to me here is to remove the `deploy.sh` call above and move all of this logic (deployment and other smoke tests) directly into `smoke-test.sh` - I am happy to do that now if you're OK with that

slimsag: Side note: It's not really clear what should be in shell commands here vs. in smoke-test.sh

davejrt: Personally, I am inclined to put the heavy duty tasks in the scripts ie: setting up volumes, testing and then execute those scripts in the shell commands. 

Previously when I've used this harness I'd been a quick and easy way to just comment out a script in shell commands if I've wanted to run the vagrant box then ssh in to manually test or tweak things before committing them. 

FYI I've moved the setup and configuration part of the test process into a new script, and the testing itself is still done with smoke-test.sh

pecigonzalo: My 2c, I would think of it the same way you think about the CI pipeline yaml. 
- Steps are listed in the yaml
- Each step is unit of work that does not know about the others

slimsag: I won't block this further because I am content with this as-is, so feel free to merge.

But the organization here still doesn't make much sense to me.

- `setup.sh` doesn't install dependencies, the `Vagrantfile` does
- The **main** property being tested by these tests is in fact the static permissions on volume directories, but we've placed that in a `setup.sh` file - why is testing done in a setup file? Should this be renamed to `perms-test.sh`?
- `smoke-test.sh` relies on the work that `setup.sh` is doing by setting up the volume directories - when this test fails in `smoke-test.sh` because a container fails to come up with e.g. wrong permissions, someone will need to find `setup.sh` and fix permissions there. This seems not super intuitive to me.
- `servers.yaml` is calling `deploy.sh` and is "responsible for starting the deployment" but `smoke-test.sh` is responsible for actually ensuring the containers came up OK?

Maybe I am missing something, but the structure that would make a lot more sense to me is:

- `servers.yaml` only calls `smoke-test.sh`, nothing else.
- everything in `setup.sh` currently goes into `smoke-test.sh`
- dependency installation goes into an external script `setup.sh` (and thus `Vagrantfile` becomes something we can just copy for other test suites in the future.)

PR #142, opened 2020-08-28
Vagrant suggestions

pecigonzalo:
I ran into a couple of issues when testing locally and ended up with this set of suggestions.

- Add more options to env-vars which can be easily added to .envrc locally
```
❯ cat .envrc
export VAGRANT_SSH_USER="foobar"
export VAGRANT_SSH_KEY="~/.ssh/google_compute_engine"
```
- Load all CI envs from files, so its not mixed. I was unsure about `machine_type` and `network` as they will likely not change
- Move user testing to our Auxiliary project
- Split base provisioning (left in the `Vagrantfile`) and test commands (in the `servers.yaml`)
- Set default ssh user to `ENV['USER']` which is what `gcloud beta compute ssh` uses and will probably match most users

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,12 @@ vagrant plugin install vagrant-google
 gcloud auth application-default login
 ```
 
-- Execute the following where `ssh_user` is a user you have GCP metadata.
+- Configure your local variables using the following environment variables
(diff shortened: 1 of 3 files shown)


davejrt:
LGTM, thanks for contributing this

PR #153, opened 2020-09-17
RFC 235: Add code intel postgres container

efritz:
A step towards [RFC 235: Move LSIF data to Postgres](https://docs.google.com/document/u/1/d/1Y9p29hK8xrPUTvBdWqP9uAHDg3JLV-HCaxsbCo9-YHQ).

Add a new copy of the postgres container to be used as a code intel-specific database. Configure the frontend to point to this new database (in addition to the current pgsql container).

Partially implements https://github.com/sourcegraph/sourcegraph/issues/13883 along with sibling PRs https://github.com/sourcegraph/sourcegraph/pull/13904 and https://github.com/sourcegraph/deploy-sourcegraph/pull/896.


deploy-codeintel-db.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
pecigonzalo, line 1: This file should be executable

efritz: Thanks for catching this 😓

pecigonzalo (approved):
LGTM, I was able to start a test deployment of docker-compose version locally.

deploy-codeintel-db.sh
```diff
@@ -0,0 +1,27 @@
...
+set -e
+
+# Description: PostgreSQL database for code intelligence data.
+#
+# Disk: 128GB / persistent SSD
+# Network: 100mbps
```
slimsag, line 7: pretty slow for the amnt of data this DB has - maybe you want 1Gbps ?

deploy.sh
```diff
@@ -25,6 +24,10 @@ for i in $(seq 0 $(($NUM_SYMBOLS - 1))); do ./deploy-symbols.sh $i; done
 for i in $(seq 0 $(($NUM_INDEXED_SEARCH - 1))); do ./deploy-zoekt-indexserver.sh $i; done
 for i in $(seq 0 $(($NUM_INDEXED_SEARCH - 1))); do ./deploy-zoekt-webserver.sh $i; done
 
+# Postgres/codeintel-db must be started before these.
+./deploy-precise-code-intel-bundle-manager.sh
+./deploy-precise-code-intel-worker.sh
```
slimsag, line 29: Please undo this change / move these back to their original position. The note below talking about Redis is describing something about file permissions (which I understand is a bit odd.)

The services themselves need to be able to start in any order, allowing for restarts if they depend on another service, and in that respect precise-code-intel should be no different than frontend. e.g. we cannot enforce startup order in Docker Compose.

Or am I missing something here?

test/pure-docker/volume-config.sh
```diff
@@ -21,6 +21,6 @@ pushd ~/sourcegraph-docker
 chown -R 100:101 gitserver* lsif-server* prometheus-v2* repo-updater* searcher* sourcegraph-frontend* symbols* zoekt*
 chown -R 999:1000 redis-store-disk redis-cache-disk
 chown -R 472:472 grafana-disk
-chown -R 999:999 pgsql-disk
+chown -R 999:999 pgsql-disk codeintel-db-disk
```
slimsag, line 24: Nothing for you to do here, but FYI if you change this code in the future it is worth calling out that this means the docs need to be updated to reflect this as mentioned in the comments above.

https://github.com/sourcegraph/sourcegraph/issues/14347#issuecomment-702338141

Did you see the comments here? (wondering if there is some way we can make it stand out more or something)

slimsag (approved):
LGTM - my only blocking comment is that you undo https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/153/files#r498454805

deploy-codeintel-db.sh
efritz: I'm not even sure where this comes from. Is this the default docker network bandwidth?

test/pure-docker/volume-config.sh
efritz: Yeah I saw, but wasn't sure what _else_ needed to be done for those migrations docs (hence the TODO you linked to).

deploy-codeintel-db.sh
slimsag: The machine it’s served on, docker and/or other container management platform restrictions (remember this deploy type is just saying “If you were to arbitrarily deploy these containers somewhere and point them out each other this is what they would need”)

For example, say someone was deploying this container and a bunch of others on a single machine with 10Gbps of network bandwidth itself - what network bandwidth can we expect to have reserved for this container?

I would put 1Gbps or 3Gbps here based on what I know about this service

deploy.sh
efritz: Undone.

deploy-codeintel-db.sh
efritz: Oh it's just a suggestion? I thought it was related to the attached network somehow.

slimsag: Just a suggestion, but someone may use this info to restrict your network speed to what you write here in the future - so..

efritz: Updated.

PR #161, opened 2020-10-07
Ignore dhall files

ElizabethStirling:
Ignore dhall files for security reasons (see [security issue 8](https://github.com/sourcegraph/security-issues/issues/8))

Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph/pull/1053) change:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ nginx/sourcegraph.key
 .vagrant
 *.log
 
- **/*.dhall 
+/right.dhall
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ nginx/sourcegraph.key
 .vagrant
 *.log
 
+ **/*.dhall 


.gitignore
```diff
@@ -8,3 +8,4 @@ nginx/sourcegraph.key
 .vagrant
 *.log
 
+ **/*.dhall
```
slimsag, line 11: @ElizabethStirling I'm not sure this PR (or the one in deploy-sourcegraph) makes sense. Dhall is just a language, akin to YAML or JSON - so ignoring _all_ Dhall files seems quite wrong to me.

I'm fine with moving forward with this if you think it will help and then later on restricting this further to a specific Dhall file name or something, though maybe we could do that now instead?

slimsag (approved):
Approving to unblock, if you feel strongly this is a net win feel free to move forward.

.gitignore
ElizabethStirling: Will update it to only ignore the single, specific file that we had issues with.

PR #171, opened 2020-11-03
tools: add new update-docker-tag.sh, add to PR checklist, update upgrade guide

bobheadxi:
Also updates the release guide to better align with https://github.com/sourcegraph/sourcegraph/pull/15281 , which closes https://github.com/sourcegraph/sourcegraph/issues/15074

### Checklist

* [x] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: https://github.com/sourcegraph/deploy-sourcegraph/pull/1203
* [x] If this change is a new service, add this service to `tools/update-docker-tags.sh`


tools/update-docker-tags.sh
```diff
@@ -0,0 +1,26 @@
...
+  -enforce="sourcegraph/precise-code-intel-bundle-manager=$CONSTRAINT" \
+  -enforce="sourcegraph/precise-code-intel-worker=$CONSTRAINT" \
+  -enforce="sourcegraph/prometheus=$CONSTRAINT" \
+  -enforce="sourcegraph/query-runner=$CONSTRAINT" \
+  -enforce="sourcegraph/redis-cache=$CONSTRAINT" \
+  -enforce="sourcegraph/redis-store=$CONSTRAINT" \
+  -enforce="sourcegraph/repo-updater=$CONSTRAINT" \
+  -enforce="sourcegraph/searcher=$CONSTRAINT" \
+  -enforce="sourcegraph/symbols=$CONSTRAINT" \
+  base/
```
slimsag, line 26: Really not a huge fan that we have to duplicate this across multiple repos now and keep it up to date. Is there a better way we could do that?

slimsag: I wonder why we cannot `-enforce="sourcegraph/.*=$CONSTRAINT"`

tools/tools.go
```diff
@@ -0,0 +1,8 @@
+//+build tools
+
+package tools
+
+import (
+	// pins update-docker-tags CLI
+	_ "github.com/slimsag/update-docker-tags"
+)
```
slimsag, line 8: Do we actually need to pin this CLI in this repo? I'm not going to change it without accounting for Sourcegraph, so it seems fine to not pin it. Given this repo is meant to be documentation/reference (see README.md) it would be nice to keep these files (tools/tools.go, go.sum, go.mod) out of this repo all together

RELEASING.md
```diff
@@ -1,20 +1,19 @@
...
-1. Find and replace `:insiders` with `:3.0.0` on all files except this one. This is required for the next step.
-2. Run these commands:
+- Tag examples: `v3.8.2`, `v3.9.2`, `customer-replica-v3.8.2`, `customer-replica-v3.9.2-2`
+- Branch examples: `3.8`, `3.9`, `3.8-customer-replica`, `3.9-customer-replica`
 
-```
-go get -u github.com/slimsag/update-docker-tags
-update-docker-tags --constraint 'sourcegraph/prometheus=<10.0' --constraint 'sourcegraph/grafana=<6.0' .
-```
+The customer replica ones are important as we must maintain some diffs for replication purposes, such as using a different Postgres version, changes to the `prometheus_targets.yml` and more. We use tags to ensure each AMI we save correlates directly with an immutable Git tag.
```
slimsag, line 10:
```suggestion
The customer replica branches are distinct living branches which have diffs we maintain for replicating some customers using the pure-docker deployment model described in the README.md of this repository (e.g. the customers' different Postgres version, changes to the `prometheus_targets.yml`, and more.) We test, tag, and release these `$VERSION-customer-replica` branches as `customer-replica-$VERSION` to produce the final diffs we send to customers running a pure-docker deployment on e.g. https://docs.sourcegraph.com/admin/updates/pure_docker
```

Rephrased to include some more detail, but also the AMI stuff won't be true soon because it's all outdated and I plan to delete them all.

RELEASING.md
```diff
@@ -1,20 +1,19 @@
...
 
-3. Confirm the diff shows the image tags being updated to the version you expect, and push directly to `master`.
+## Releasing a new version
 
-## 2) Create the release branch:
+### Create the release branch
+
+> ⚠️ If you are using the Sourcegraph release tooling, this step will be done for you in the PR it creates.
```
slimsag, line 16: What is the "Sourcegraph release tooling" and where do I find it? Can we link to it here, or a handbook page, etc.?

RELEASING.md
```diff
@@ -23,7 +22,20 @@ git checkout -B 3.17
...
+
+In the latest release branch you created:
+
+1. Run `tools/update-docker-tags.sh $VERSION`
+2. Confirm the diff shows the image tags being updated to the version you expect, and push directly to the release branch.
+
+### Cherry-pick image update to master
+
+Cherry-pick the commit where you update images on the release branch back into `master`.
```
slimsag, line 36: so the release tooling will make the PR to update images in only the release branch? is there a reason we can't invert this and have it do it against `master`, and then create the release branch from that commit in `master`?

RELEASING.md
```diff
@@ -107,27 +119,12 @@ Replace the old version with the new version in the `master` branches of the fol
...
-Write a message to #dev-announce:
-
-> Docker Compose v3.9.2 has been released.
-
-## 9) Message @stephen on Slack
+### Message @stephen on Slack
 
 > I am releasing deploy-sourcegraph-docker, please release pure-docker.
 
 If you are Stephen, follow [releasing pure-docker](#releasing-pure-docker) below.
```
slimsag, line 126: I forgot this was still here, will send a PR later to remove it.

RELEASING.md
```diff
@@ -107,27 +119,12 @@ Replace the old version with the new version in the `master` branches of the fol
...
     - Update `latestReleaseDockerComposeOrPureDocker` in https://github.com/sourcegraph/sourcegraph/blob/master/cmd/frontend/internal/app/pkg/updatecheck/handler.go#L47
 
-## 8) Post that you are done.
-
-Write a message to #dev-announce:
-
-> Docker Compose v3.9.2 has been released.
-
-## 9) Message @stephen on Slack
+### Message @stephen on Slack
```
slimsag: I went ahead and removed this section calling out to me to follow these steps, since others (primarily Uwe and Dave) will now be doing this. If you could update the upstream release checklist to include a step to consider doing a pure-docker release (e.g. in the case of a major release, but not strictly all patch releases) that'd be excellent!

slimsag (approved):
I left a lot of feedback inline, LGTM otherwise. No need for re-review from me.

tools/tools.go
bobheadxi: @davejrt was unable to update the tags through the release tool during the last release because his version of this tool did not have my patch adding `-enforce` - I think it is generally good practice to have managing tool versions be a bit seamless where possible, since even if there is documentation, being faced with an error that's related to an out-of-date tool is usually the most frustrating experience ever 😓

RELEASING.md
bobheadxi: I just felt it aligned better with the `deploy-sourcegraph` scheme, and I've been trying to have the tool create "final" PRs - e.g. PRs where the merge _is_ the final state of the release (thus, the release campaign becomes a good snapshot of the state of a release).

IMO updating `master` is a secondary concern by comparison (since nobody should be using it), so if a delayed manual step must happen I'd rather it be the `master` update than the release branch update

tools/update-docker-tags.sh
bobheadxi: > Really not a huge fan that we have to duplicate this across multiple repos now and keep it up to date. Is there a better way we could do that?

Yeah, the duplication was unexpected (didn't realize this tool was used in this repo as well) - I think the best home for this script in the future would be in `sourcegraph/sourcegraph`. If we have a "master" list of services somewhere, could generate this script using that as well.

> I wonder why we cannot -enforce="sourcegraph/.*=$CONSTRAINT"

I attempted to add this feature when working on https://github.com/sourcegraph/deploy-sourcegraph/pull/857 , but was unable to:

> Images are defined by hand - I tried an approach with a more generic approach to update all Sourcegraph images, but we have some versioning inconsistencies that made things a bit difficult, so I gave up on that (https://github.com/bobheadlabs/update-docker-tags/commit/38ab3ab34faa97fa18b80a4d3ae76bde9c3d14e9)

I think most notably `sourcegraph/postgres`, etc. had non-`3.x.y` tags which would immediately break a feature like this. Attempts to get around it with regex was extremely painful and I gave up 🙃 Could revisit, though, but I think a better way of maintaining the list of services is better.

Overall, I think the overhead is (hopefully) low, given service additions and removals happen rarely (I think?)

RELEASING.md
bobheadxi: Added links in 5a5046e1359f447d0d08e6fd691f79496e69b894 !

bobheadxi:
Thanks for the review @slimsag - merging now, but I left replies to your questions/concerns, happy to discuss further if needed!

tools/update-docker-tags.sh
pecigonzalo: What about making it a `buildkite` plugin?

RELEASING.md
bobheadxi: Noted here: https://github.com/sourcegraph/sourcegraph/pull/15281/commits/d8bcb6a9a8aa822b55c7fee1a3c2e58f8fb2d92b

tools/update-docker-tags.sh
bobheadxi: This isn't used in buildkite - the one place where it runs in hosted automation is GitHub Actions, and in all other cases it's run on somebody's dev machine (including in the release tooling)

RELEASING.md
slimsag: My concern here is that someone will forget to cherry-pick the commit. But it sounds like you have grander plans here and are aware of that, so I'm happy :)

PR #186, opened 2020-11-26
qa: add logging

davejrt:
add some additional logging to our docker tests to help troubleshoot issues in CI


test/Vagrantfile
```diff
@@ -8,6 +8,10 @@ require 'yaml'
 
 # Read YAML file with box details
 servers = YAML.load_file('servers.yaml')
+
+# Base image to use for test instances
+image = 'e2e-491e9f27-2020-10-28t17-32-00z'
```
pecigonzalo, line 13: It might be good to pass this from `servers.yaml` or env-var

test/docker-compose/smoke-test.sh
```diff
@@ -23,12 +22,13 @@ fi
...
+	containers_running=$(echo "$containers" | grep -c "running")
 	if [ "$containers_running" -ne "$expect_containers" ]; then
-		docker ps
 		echo
-		echo "TEST FAILURE: expected $expect_containers containers running, found $containers_running"
-    echo containers
+		containers_failing=$(docker ps --format '{{.Names}}:{{.Status}}' | grep -v Up | cut -f 1 -d :)
+		echo "TEST FAILURE: expected $expect_containers containers running, found $containers_running. The following containers are failing: $containers_failing"
+		echo ""
+		for cf in $containers_failing; do docker logs "$cf" >/deploy-sourcegraph-docker/"$cf".log; done
```
pecigonzalo, line 31:
```suggestion
		for cf in $containers_failing; do docker logs -t "$cf" >/deploy-sourcegraph-docker/"$cf".log; done
```
Show timestamp

test/docker-compose/smoke-test.sh
```diff
@@ -23,12 +22,13 @@ fi
 echo "TEST: Checking every 10s that containers are running for 5 minutes..."
 for i in {0..30}; do
 	containers=$(docker ps --format '{{.Names}}' | xargs -I{} -n1 sh -c "printf '{}: ' && docker inspect --format '{{.State.Status}}' {}")
-	containers_running=$(echo "$containers" | grep "running" | wc -l)
+	containers_running=$(echo "$containers" | grep -c "running")
```
pecigonzalo, line 25: Wouldnt this make `[ "$containers_running" -ne "$expect_containers" ]` fail? As it expects a count.

PS: I would change all `[]` to `[[]]`

test/pure-docker/smoke-test.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -eufxo pipefail
+set -eufo pipefail
 
 cd /deploy-sourcegraph-docker
 sudo su
```
pecigonzalo, line 5: Should we remove this from this one as well?

test/pure-docker/smoke-test.sh
```diff
@@ -21,24 +21,26 @@ sleep 10
...
-    echo "Containers running OK.. waiting 10s"
-    sleep 10
+	containers=$(docker ps --format '{{.Names}}' | xargs -I{} -n1 sh -c "printf '{}: ' && docker inspect --format '{{.State.Status}}' {}")
+	containers_running=$(echo "$containers" | grep -c "running")
+	if [ "$containers_running" -ne "$expect_containers" ]; then
+		echo
+		containers_failing=$(docker ps --format '{{.Names}}:{{.Status}}' | grep -v Up | cut -f 1 -d :)
+		echo "TEST FAILURE: expected $expect_containers containers running, found $containers_running. The following containers are failing: $containers_failing"
+		echo ""
+		for cf in $containers_failing; do docker logs "$cf" >/deploy-sourcegraph-docker/"$cf".log; done
```
pecigonzalo, line 39:
```suggestion
		for cf in $containers_failing; do docker logs -t "$cf" >/deploy-sourcegraph-docker/"$cf".log; done
```

test/pure-docker/smoke-test.sh
```diff
@@ -21,24 +21,26 @@ sleep 10
...
-    if [ "$containers_running" -ne "$expect_containers" ]; then
-        docker ps
-        echo
-        echo "TEST FAILURE: expected $expect_containers containers running, found $containers_running"
-        exit 1
-    fi
-    echo "Containers running OK.. waiting 10s"
-    sleep 10
+	containers=$(docker ps --format '{{.Names}}' | xargs -I{} -n1 sh -c "printf '{}: ' && docker inspect --format '{{.State.Status}}' {}")
+	containers_running=$(echo "$containers" | grep -c "running")
```
pecigonzalo, line 33: Similar to the other file, I think this will make it fail the `if` check.

pecigonzalo (changes requested):
I notice a lot of the code is repeated between the tests, maybe we can extract that to a shared file.

test/docker-compose/smoke-test.sh
davejrt:
```
  -c, --count               print only a count of selected lines per FILE
```

davejrt:
@pecigonzalo can you take a look at this refactor and let me know what you think?

pecigonzalo (approved):
I was thinking more in the lines on the lines of keeping 2 scripts for setup, but one with what you did in `test_count` and `test_containers` as its the part that is reused, and could take advantage of the `shell_commands` part to choose what to run for each type of test.

davejrt:
Got it, I guess I went down this path because everything is in one file, which means there's less digging through different scripts trying to work out what's being called where. 

I'm going to merge this because I think it solves our immediate concern of not having visibility to the errors we're getting. 

But I am happy to iterate over this more to find the sweet spot.

PR #195, opened 2020-12-20
fix path in tools/update-docker-tags.sh

uwedeportivo:


uwedeportivo:
i'll cherry-pick it into 3.23

bobheadxi (approved):

PR #213, opened 2021-01-10
cadvisor: add commented support for macos

bobheadxi:
closes https://github.com/sourcegraph/sourcegraph/issues/17072, see https://github.com/sourcegraph/sourcegraph/issues/17072#issuecomment-757460360

### Checklist


daxmc99 (approved):
LGTM! Mounting the docker socket in this way might give the containers `ro` access to the host but I think that tradeoff is acceptable here.

uwedeportivo (approved):

PR #256, opened 2021-02-11
prometheus job label

uwedeportivo:
Fixes https://github.com/sourcegraph/sourcegraph/issues/16552 for docker compose, pure docker.

I'm not clear what the previous value achieved. No rewrite rules kicked in and the label value in a live environment is "node" for all the containers and services. Don't think that's too useful. I left a couple as "node" where I couldn't find k8s or single server counterparts.

### DIFF
--- a/prometheus/prometheus_targets.yml
+++ b/prometheus/prometheus_targets.yml
@@ -7,16 +7,48 @@
     #- codeintel-db:9187
     #- redis-cache:9121
     #- redis-store:9121
+- labels:
+    job: node
+  targets:
     - cadvisor:8080
-    - sourcegraph-frontend-internal:6060
+- labels:
+    job: github-proxy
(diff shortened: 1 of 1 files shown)


slimsag:
Hmmm, previously these metrics showed up in Grafana as having labels like `sourcegraph-frontend-internal:6060` just based on the `target` here - I wonder what changed that made that no longer work?

uwedeportivo:
@slimsag the value of the `instance` label is something like `sourcegraph-frontend-internal:6060`, so that feels correct. not sure if we use the `job` label much because if we do, somebody would have screamed by now.

image

slimsag:
Ahhh, okay, that makes sense. Thanks for that clarification!

slimsag (approved):

PR #258, opened 2021-02-16
migrate single server to compose

davejrt:
A script to migrate Sourcegraph single server deployments to compose:

I chose editing the dump file for sql commands as the bash foo to escape quotes was awful. Happy to revisit this if someone disagrees but I feel as though I burnt too much time here before it occurred to me it'd just be easier do it this way. 

fixes https://github.com/sourcegraph/sourcegraph/issues/11600

### DIFF
--- a/tools/migrate.sh
+++ b/tools/migrate.sh
@@ -1,39 +1,53 @@
 #!/bin/bash
 
-set -euxo pipefail
+set -uxo pipefail
 
 # shellcheck source=migration_vars
 source migration_vars
 
 IMAGE=${IMAGE:-sourcegraph/server}
-# trap 'rm -f ./*pg.out' EXIT
+sg_dump_file=$(mktemp)
+codeintel_dump_file=$(mktemp)
 
+trap 'rm -f ${sg_dump_file} ${codeintel_dump_file}' EXIT
+
+# Get the  ID of the sourcegraph container
 CONTAINER_ID=$(ssh "${SRC_HOST}" docker ps | grep "${IMAGE}:${VERSION}" | cut -f 1 -d ' ')
 
-ssh -t "${SRC_HOST}" "docker exec ${CONTAINER_ID} sh -c 'pg_dumpall --verbose --username=postgres' " >sourcegraph_pgall.out
-# ssh -t "${SRC_HOST}" "docker exec ${CONTAINER_ID} sh -c 'pg_dump --verbose --username=postgres sourcegraph-codeintel' " >codeintel_pg.out
+# Dump sourcegraph db and alter script for migration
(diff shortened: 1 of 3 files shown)


tools/migrate.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+# Please ensure the variales in migration vars have been assigned before running this script.
```
uwedeportivo, line 3:
```suggestion
# Please ensure the variables in migration vars have been assigned before running this script. 
```

tools/migrate.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+# Please ensure the variales in migration vars have been assigned before running this script. 
+
+set -euxo pipefail
+
+# shellcheck source=migration_vars
+source migration_vars
```
uwedeportivo, line 8: maybe check here that the variables you assume as set are actually set

uwedeportivo (approved):

tools/migrate.sh
davejrt: OK yeah, in testing I never bothered to check for this and assumed `-u` would pick that up but it doesn't work if you're sourcing the file.

davejrt: Running the `source <file>` in a subshell fixes this though. New commit up now

tools/migrate.sh
```diff
@@ -0,0 +1,62 @@
...
+echo "Obtaing sourcegraph container ID"
+container_id=$(ssh "${src_host}" docker ps | grep "${image}:${version}" | cut -f 1 -d ' ')
+
+# Dump sourcegraph db and alter script for migration
+echo "Dumping sourcegraph database"
+ssh ${src_host} "docker exec ${container_id} pg_dump -C --username=postgres sourcegraph" >${sg_dump_file}
+sed -i '/^CREATE DATABASE/a CREATE USER postgres WITH SUPERUSER CREATEROLE CREATEDB REPLICATION BYPASSRLS;' ${sg_dump_file} >/dev/null 2>&1
+printf '\connect -reuse-previous=on dbname=postgres\nDROP DATABASE sg;\nALTER DATABASE "sourcegraph" RENAME TO sg;\nALTER DATABASE sg OWNER TO sg;\n' >>${sg_dump_file}
+
+# Dump codeintel db and alter script for migration
```
daxmc99, line 28: cc @slimsag, is codeinsights supported in our `server` image?
Not sure if we need to add migration steps here as well

slimsag: It is not supported. Deploying TimescaleDB inside of single-container would take a fair amount of effort, and given it is for demo purposes only we have made the decision for now to not support it. (a more "minimal" version, without any TimescaleDB backend, might work - though)

PR #274, opened 2021-03-01
use healthcheck to fix db lock

davejrt:
This stops the frontend containers competing for a lock and erroring during the migrations. 

This ensures the internal pod comes up first, as other services start to talk to this before an user would start accessing the UI. In addition, if a customer wants to create multiple replicas of the frontend the start order and potential conflict between containers is still accounted for.

### DIFF
--- a/deploy-frontend-internal.sh
+++ b/deploy-frontend-internal.sh
@@ -18,6 +18,11 @@ docker run --detach \
     --restart=always \
     --cpus=4 \
     --memory=8g \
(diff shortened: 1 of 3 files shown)


daxmc99 (approved):
This may also fix https://github.com/sourcegraph/customer/issues/208 
I didn't realize we had two frontends on docker-compose that pointed towards the same image... it is not clear to me what the split is here.

PR #276, opened 2021-03-02
use healthchecks in pure docker

davejrt:


pecigonzalo (approved):
LGTM, we could also move this to the `Dockerfile` definition for docker/docker-compose.

davejrt:
Throwing in a random number for the server name too, as there is an edge case where they both get started at the exact same time and one box fails because of it

eg: https://buildkite.com/sourcegraph/deploy-sourcegraph-docker/builds/3452#9fbc1493-8c59-47ca-861f-63c18fa709d4

davejrt:
> LGTM, we could also move this to the `Dockerfile` definition for docker/docker-compose.

How would this impact the same image running in kubernetes? Given we don't differentiate at all.

pecigonzalo:
I dont think Kubernetes uses it, so it should not but maybe something to check later given this fix is tested.

PR #277, opened 2021-03-02
increase retries for frontend

davejrt:
Still getting some failures, adjusting this for a little more leeway when bringing up frontend containers.


daxmc99 (approved):
Should we also bump up the retries on the internal frontend?

davejrt:
> Should we also bump up the retries on the internal frontend?

Yeah, for the sake of consistency I'll dump them in all frontends. Good call, thanks :-)

PR #320, opened 2021-03-30
Use Postgres 12

efritz:
### Checklist

* [ ] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: https://github.com/sourcegraph/deploy-sourcegraph/pull/2193

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -95,7 +95,7 @@ At this point you should evaluate the `git status` output as well as all the cha
 During this process you will run into two merge conflicts:
 
 - Do not commit: `deploy-caddy.sh` or changes related to it, as `deploy-apache.sh` is used here.
-- Do not commit: changes to `deploy-pgsql.sh`, as Postgres 12.6 is used here.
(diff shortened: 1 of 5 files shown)


RELEASING.md
```diff
@@ -95,7 +95,7 @@ At this point you should evaluate the `git status` output as well as all the cha
 During this process you will run into two merge conflicts:
 
 - Do not commit: `deploy-caddy.sh` or changes related to it, as `deploy-apache.sh` is used here.
-- Do not commit: changes to `deploy-pgsql.sh`, as Postgres 9.6 is used here.
+- Do not commit: changes to `deploy-pgsql.sh`, as Postgres 12.6 is used here.
```
davejrt, line 98: This is actually correct as it is...it refers to a specific customer deployment in this case.

davejrt:
```suggestion
- Do not commit: changes to `deploy-pgsql.sh`, as Postgres 9.6 is used here.
```

davejrt (approved):

PR #451, opened 2021-06-22
update image used by vagrant

davejrt:
Uses image built from https://github.com/sourcegraph/infrastructure/pull/2649 with sudo access for the vagrant user

Aims to resolve https://github.com/sourcegraph/sourcegraph/issues/21594


davejrt:
Merging as the [sister change](https://github.com/sourcegraph/sourcegraph/pull/22294) was approved and has had no ill effects on other pipelies

PR #489, opened 2021-08-03
Fix Jaeger Tracing in Docker Compose

abeatrix:
Jaeger is not working on Docker Compose currently due to missing env var stated in https://github.com/sourcegraph/sourcegraph/issues/21677#issuecomment-889433798
This PR adds the missing env var SAMPLING_STRATEGIES_FILE=/etc/jaeger/sampling_strategies.json so that jaeger can work on Docker Compose instances properly. 

### Checklist

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -502,6 +502,8 @@ services:
       - sourcegraph
     restart: always
     command: ['--memory.max-traces=20000']
+    environment:
+      - 'SAMPLING_STRATEGIES_FILE=/etc/jaeger/sampling_strategies.json'
 
   # Description: PostgreSQL database for various data.
   #


bobheadxi (approved):

bobheadxi:
Seemed to pass after giving the CI a kick, seemed to be a network flake :)

abeatrix:
Thanks for checking @bobheadxi ! Will I need to update the changelog to reflect the changes as well?

bobheadxi:
I would recommend something like this:

> ### Changed
> 
> - docker-compose: the Jaeger container's `SAMPLING_STRATEGIES_FILE` now has a default value. If you are currently using a custom sampling strategies configuration, you may need to make sure your configuration is not overridden by the change when upgrading.

PR #580, opened 2021-09-09
prometheus: scrape pg, redis instances by default

bobheadxi:
Convenience and parity with k8s (would have been useful for https://github.com/sourcegraph/customer/issues/470 where I realized too late they were on Compose)

Without this the postgres dashboards are also super unhelpful on Compose, e.g. https://demo.sourcegraph.com/-/debug/grafana/d/postgres/postgres?orgId=1 

### Checklist

### DIFF
--- a/prometheus/prometheus_targets.yml
+++ b/prometheus/prometheus_targets.yml
@@ -2,15 +2,6 @@
 - labels:
     job: node
   targets:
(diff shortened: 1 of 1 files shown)


daxmc99 (approved):
LGTM. I honestly didn't know that we didn't monitor these services in docker-compose 🤯  
This probably warrants a changelog entry though

PR #684, opened 2021-12-23
Add in migrator containers

danieldides:
Part of: https://github.com/sourcegraph/sourcegraph/issues/25256. Closes https://github.com/sourcegraph/sourcegraph/issues/31435.

This PR will add in the new migrator containers. The migrator containers define:

```
    depends_on:
      pgsql:
        condition: service_healthy
```
to wait for the database instances to start up, and then the `frontend-internal` service is altered to define:

```
      migrator-frontend:
        condition: service_completed_successfully
      migrator-codeintel:
        condition: service_completed_successfully
```
to wait for the migrations to finish executing. Additionally, the `frontend-internal` service now sets the environment variable `'NEW_MIGRATIONS=true'` to prevent that service from executing the migrations itself.


docker-compose/docker-compose.yaml
```diff
@@ -145,11 +212,10 @@ services:
       retries: 5
       start_period: 300s
     depends_on:
-      pgsql:
-        condition: service_healthy
-      codeintel-db:
-        condition: service_healthy
-
+      migrator-frontend:
+        condition: service_completed_successfully
```
efritz, line 216: Awesome.

efritz (approved):

docker-compose/docker-compose.yaml
```diff
@@ -1,5 +1,53 @@
...
+      - CODEINSIGHTS_PGDATABASE=postgres
+      - CODEINSIGHTS_PGSSLMODE=disable
+    restart: "on-failure"
+    networks: 
+      - sourcegraph
+    depends_on:
+      pgsql:
+        condition: service_healthy
+      codeintel-db:
+        condition: service_healthy
```
caugustus-sourcegraph, line 49: This should also depend on codeinsights-db, right?

docker-compose/docker-compose.yaml
```diff
@@ -90,6 +138,7 @@ services:
       - 'JAEGER_SERVER_URL=http://jaeger:16686'
       - 'GITHUB_BASE_URL=http://github-proxy:3180'
       - 'PROMETHEUS_URL=http://prometheus:9090'
+      - 'NEW_MIGRATIONS=true'
```
caugustus-sourcegraph, line 141: Is this env variable still relevant for the release?

docker-compose/docker-compose.yaml
```diff
@@ -172,6 +219,9 @@ services:
       - sourcegraph
     restart: always
     hostname: gitserver-0
+    depends_on:
```
caugustus-sourcegraph, line 222: What's the reasoning behind adding all these new `depends_on` settings? I'm not 100% clear on the dependency order but is it possible we're getting in a loop here, where frontend-internal can't be healthy until some of these services are healthy, and vice-versa?

Regardless, I don't think the dependency should be added for Prometheus/Grafana since those are definitely unrelated to frontend-internal.

https://docs.sourcegraph.com/dev/background-information/architecture

docker-compose/docker-compose.yaml
```diff
@@ -1,5 +1,53 @@
...
+      - PGSSLMODE=disable
+
+      - CODEINTEL_PGHOST=codeintel-db
+      - CODEINTEL_PGPORT=5432
+      - CODEINTEL_PGUSER=sg
+      - CODEINTEL_PGPASSWORD=sg
+      - CODEINTEL_PGDATABASE=sg
+      - CODEINTEL_PGSSLMODE=disable
+
+      - CODEINSIGHTS_PGHOST=codeinsights-db
```
caugustus-sourcegraph, line 36: (Not really relevant to this PR) Are there any plans to update the frontend `CODEINSIGHTS_PGDATASOURCE` connection string to follow this same pattern? It's less relevant for docker deployments but the difference in format makes secure credential management difficult to manage.

docker-compose/docker-compose.yaml
```diff
@@ -1,5 +1,53 @@
...
+  # Ports: None
+  #
+  # This container should start after the database instance starts, perform the migrations
+  # and then exit cleanly. After exiting the `frontend` service is allowed to start.
+  #
+  # If you're using a DB instance hosted outside of docker-compose, the environment variables
+  # for this container will need to be updated to reflect the new connection information.
+  migrator:
+    container_name: migrator
+    image: 'index.docker.io/sourcegraph/migrator:122149_2021-12-17_1f7179c'
```
caugustus-sourcegraph, line 16: This new image should be added to the [list of images here](https://github.com/sourcegraph/sourcegraph/blob/main/enterprise/dev/ci/images/images.go#L58) so it can get auto-updated by Renovate and the release tooling.

docker-compose/docker-compose.yaml
danieldides: `codeinsights-db` unfortunately doesn't expose a healthcheck endpoint defined that we're able to check against so we can't wait for the `service_healthy` condition. I looked into adding it but because it's leveraging TimescaleDB I'm a little weary on how to check if it's actually "ready".

docker-compose/docker-compose.yaml
danieldides: Good catch, this won't be required come release.

docker-compose/docker-compose.yaml
danieldides: I'm honestly not sure. I agree, it's a little odd having it be different. `codeinsights` also doesn't use the same `sg`/`sg` pattern that the other DB's use.

We had a brief discussion about it on slack here: https://sourcegraph.slack.com/archives/C02FPB08MN3/p1642189372001800?thread_ts=1642106589.009300&cid=C02FPB08MN3

docker-compose/docker-compose.yaml
caugustus-sourcegraph: Boo! That ruins the totally-awesome suggestion I was going to make for the docs PR as an alternative to the manual docker run command. You can run `docker-compose up migrator` and it only starts up the migrator container and associated dependencies.

But, without codeinsights-db, the migrator can't run so my suggestion won't work 😢

danieldides:
@caugustus-sourcegraph  / @daxmc99 / @kevinwojo I've removed the `depends_on` conditions from most of the containers. I've tested it on my local machine and it seems to work exactly as we expected -- lots of errors on startup but once the migrations are complete and the frontend stabilizes, it's good to go. Do any of you have any additional changes or tests that you think need to be done to this before we close it? I can't quite remember from the call if we decided to enable `NEW_MIGRATIONS` as well.

docker-compose/docker-compose.yaml
```diff
@@ -1,5 +1,56 @@
...
+  #
+  # This container should start after the database instance starts, perform the migrations
+  # and then exit cleanly. After exiting the `frontend` service is allowed to start.
+  #
+  # If you're using a DB instance hosted outside of docker-compose, the environment variables
+  # for this container will need to be updated to reflect the new connection information.
+  migrator:
+    container_name: migrator
+    image: 'index.docker.io/sourcegraph/migrator:3.35.1'
+    cpus: 4
```
caugustus-sourcegraph, line 17: The cpu/mem limit here seems really high for a migrator container - can we figure out a better benchmark?

docker-compose/docker-compose.yaml
```diff
@@ -90,6 +141,7 @@ services:
       - 'JAEGER_SERVER_URL=http://jaeger:16686'
       - 'GITHUB_BASE_URL=http://github-proxy:3180'
       - 'PROMETHEUS_URL=http://prometheus:9090'
+      - 'NEW_MIGRATIONS=true'
```
caugustus-sourcegraph, line 144: These NEW_MIGRATIONS flags should be removed, right? We'll use them for testing until https://github.com/sourcegraph/sourcegraph/pull/29144 gets merged the but they shouldn't be included in this PR.

docker-compose/docker-compose.yaml
```diff
@@ -1,5 +1,56 @@
...
+  # Ports: None
+  #
+  # This container should start after the database instance starts, perform the migrations
+  # and then exit cleanly. After exiting the `frontend` service is allowed to start.
+  #
+  # If you're using a DB instance hosted outside of docker-compose, the environment variables
+  # for this container will need to be updated to reflect the new connection information.
+  migrator:
+    container_name: migrator
+    image: 'index.docker.io/sourcegraph/migrator:3.35.1'
```
caugustus-sourcegraph, line 16: Reminder that we should bump this image version at some point before merging

docker-compose/docker-compose.yaml
```diff
@@ -144,12 +192,6 @@ services:
       timeout: 10s
       retries: 5
       start_period: 300s
-    depends_on:
```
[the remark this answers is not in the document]
danieldides: Did we bump our minimum docker-compose version requirement? This `depends_on` I believe had to be removed because `service_completed_successfully` was unavailable in our old compose version.

docker-compose/docker-compose.yaml
danieldides: I copied these specs from the `frontend` containers with the expectation that once these migrations are done, the container terminates and should release the resources. We can lower them dramatically, the migrator needs very little resources to issue the SQL, the real lift is on the DB itself.

docker-compose/docker-compose.yaml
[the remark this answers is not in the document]
caugustus-sourcegraph: Right, we can't use `depends_on`: with a `condition: service_completed_successfully` without upgrading docker-compose. But we can keep the existing `condition: service_healthy` entries

docker-compose/docker-compose.yaml
danieldides: Bumped!

docker-compose/docker-compose.yaml
danieldides: Correct. Removed!

docker-compose/docker-compose.yaml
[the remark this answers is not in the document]
danieldides: Doh, I was looking at the conversation view and though we were talking about a different part of the changes. I'll add these back in now. 🤦‍♀️

caugustus-sourcegraph (approved):

docker-compose/docker-compose.yaml
```diff
@@ -102,7 +154,9 @@ services:
       - sourcegraph
     restart: always
     depends_on:
-      sourcegraph-frontend-internal:
```
caugustus-sourcegraph, line 105: Why did this change end up being necessary? Removing the `sourcegraph-frontend-internal` means that we're going to restart both frontend containers at the same time during upgrdaes. That could be more disruptive if there's a problem bringing up the frontend service, running migrations, or any of its unofficial dependencies (gitserver etc.) are slow to start.

With the dependency on `sourcegraph-frontend-internal`, we keep the old public-facing frontend running until the internal frontend is healthy, and then swap over, to reduce the impact on users. It's unexpected that adding a migrator would require a change to this setup.

docker-compose/docker-compose.yaml
caugustus-sourcegraph: Reminder that this change snuck back in during CI testing and should be removed

docker-compose/docker-compose.yaml
danieldides: Ultimately Dave and I didn't figure out why this was happening but when the `sourcegraph-frontend` was waiting for `sourcegraph-frontend-internal` it would never come up. `-internal` would prompt saying it's available on whatever port but the `frontend` container would never start up despite having a `restart: always` configuration which would eventually result in the container getting marked as unhealthy and tests failing. It was happening both locally and in the testing environment.

caugustus-sourcegraph: Hm - I think what's happening is that now the `sourcegraph-frontend` container is failing its healthcheck (and restarting?) before the migrator starts. Both those containers start up at virtually the same moment. Once the migrator container begins running, the frontend healthcheck seems to notice that.

If I increase the interval or retries count on the `sourcegraph-frontend` healthcheck, `sourcegraph-frontend-internal` successfully starts up. Not great, though. Maybe we are just stuck with removing this dependency until we can upgrade docker-compose.

@efritz - does my theory sound plausible?

caugustus-sourcegraph: @danieldides - With the `sourcegraph-frontend-internal` dependency removed, a failed migration now causes downtime. I was reluctant to tweak the interval / retries count but it may be a better option, compared to affecting customers. What do you think?

@efritz - It might help! But I'm not certain, and theoretically this problem will go away when we upgrade docker-compose anyway.

PR #716, opened 2022-01-21
Update RELEASING.md

caugustus-sourcegraph:
### Checklist


kevinwojo (approved):
Thanks for the reminder!

caugustus-sourcegraph:
😄  This is definitely a hack - it'd be better to actually fix whatever is causing this problem, but it was way easier to just say "look at this thing".

PR #724, opened 2022-01-26
Update default pgsql docker image for pure-docker, process changes

caugustus-sourcegraph:
I realized that the reason the pgsql version for pure-docker releases never updated is that the initial batch change wasn't updating the version. The container image in master was never updated to the new name, so the update-tags script couldn't find it.

I also included a small usability change on the pure-docker update process, so we don't have to resolve merge conflicts. In the next stage, we visually inspect the diffs and revert any unnecessary changes so this should be safe.

### Checklist


kevinwojo (approved):
Nice catch!

PR #748, opened 2022-02-03
Update deploy-pgsql.sh

caugustus-sourcegraph:
Fix incorrect version on pgsql

### Checklist


michaellzc (commented):
Emm, I don't remember this file showed up in the diff earlier, or maybe I missed something

michaellzc (approved):

caugustus-sourcegraph:
Yeah, deploy-pgsql.sh hasn't been getting updated in the batch change so we've had to manually correct it a few times. This was _supposed_ to be fixed by https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/724, so I need to look into what's going on there.

caugustus-sourcegraph:
I'm going to merge this then rebase/force-update the customer-replica branch so we get a single diff.

PR #750, opened 2022-02-04
Use cpu_shares

daxmc99:
Switch to cpu_shares with 1cpu == 1024 base scheduling priority.
This should solve issues with needing to manually scale services to
fully utilize the host.

Fixes of https://github.com/sourcegraph/sourcegraph/issues/30214

### Checklist

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -18,7 +18,7 @@ services:
   caddy:
     container_name: caddy
     image: 'index.docker.io/caddy/caddy:alpine@sha256:83d9aa7a5f1bbcc0fc1b4720c183a5ec53dae7dc5d9fa555daf3db345010e7f9'
-    cpus: 4
+    cpu_shares: 4096
     mem_limit: '4g'
     environment:
       - 'XDG_DATA_HOME=/caddy-storage/data'
@@ -70,7 +70,7 @@ services:
   sourcegraph-frontend-0:
     container_name: sourcegraph-frontend-0
     image: 'index.docker.io/sourcegraph/frontend:insiders@sha256:7c453e79a63e92b79901a6b523859d43e870f978162a60a4b219a6e54aa6c98a'
-    cpus: 4
+    cpu_shares: 4096
     mem_limit: '8g'
     environment:
       - DEPLOY_TYPE=docker-compose
@@ -114,7 +114,7 @@ services:
   sourcegraph-frontend-internal:
     container_name: sourcegraph-frontend-internal
     image: 'index.docker.io/sourcegraph/frontend:insiders@sha256:7c453e79a63e92b79901a6b523859d43e870f978162a60a4b219a6e54aa6c98a'
-    cpus: 4
+    cpu_shares: 4096
     mem_limit: '8g'
     environment:
       - DEPLOY_TYPE=docker-compose
@@ -160,7 +160,7 @@ services:
   gitserver-0:
     container_name: gitserver-0
     image: 'index.docker.io/sourcegraph/gitserver:insiders@sha256:4f67f43a5bbefef5829da08be8f96235935d048218a22b96dd9f8af6f7fe7256'
-    cpus: 4
+    cpu_shares: 4096
     mem_limit: '8g'
     environment:
       - GOMAXPROCS=4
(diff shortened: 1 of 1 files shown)


daxmc99:
Reasons we hesitated to do this in the past were around the concern that this would allow extremely small hosts to start instead of blocking on the number of CPUs. 

Maybe a mix would be a good idea here? Regardless, we are facing scaling issues ourselves because of this.

docker-compose/docker-compose.yaml
```diff
@@ -18,7 +18,7 @@ services:
   caddy:
     container_name: caddy
     image: 'index.docker.io/caddy/caddy:alpine@sha256:83d9aa7a5f1bbcc0fc1b4720c183a5ec53dae7dc5d9fa555daf3db345010e7f9'
-    cpus: 4
```
caugustus-sourcegraph, line 21: What's the impact of using cpu_shares on our Grafana metrics / monitoring?

I'd be interested in trying this out on a managed instance first, then adopting it as the standard.

docker-compose/docker-compose.yaml
```diff
@@ -160,7 +160,7 @@ services:
   gitserver-0:
     container_name: gitserver-0
     image: 'index.docker.io/sourcegraph/gitserver:insiders@sha256:4f67f43a5bbefef5829da08be8f96235935d048218a22b96dd9f8af6f7fe7256'
-    cpus: 4
+    cpu_shares: 4096
     mem_limit: '8g'
     environment:
       - GOMAXPROCS=4
```
caugustus-sourcegraph, line 166: Do we need to review the GOMAXPROCS settings, or does everything just work out like before? Should we remove these env var's entirely?

daxmc99: I suggest we remove these entirely
cc @sourcegraph/repo-management

caugustus-sourcegraph: To be extra clear (since repo-management is tagged here): we set GOMAXPROCS on all of the Sourcegraph-proper services (frontend, gitserver, etc.), and we're proposing that we remove that setting from all the services.

daxmc99: Nothing like tripping over a fundamental language Golang issue in a Github comment: https://github.com/golang/go/issues/33803 
But for this case, yes, we want to drop these everywhere. If the CPU isn't being actively used by a higher priority process it should be allowed to use all cores.

kevinwojo: An issue opened almost 3 years ago in golang... woof!

daxmc99: https://github.com/uber-go/automaxprocs might work here. 
Before merging we should test this on devmanaged

docker-compose/docker-compose.yaml
daxmc99: Shouldn't have any impact. Our cAdvisor metrics look at CPU usage. They might compare it against cpu_limit though which could be an issue

daxmc99: https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md & ctrl-f limits. 
There is no `cpu_limits` only CFS periods so I highly doubt any cAdvisor issues would be affected by this

docker-compose/docker-compose.yaml
caugustus-sourcegraph: Deployed to devmanaged, and it looks like this does affect some cadvisor CPU metrics. Most of the charts look ok but this one is [missing CPU data since the update](https://devmanaged.sourcegraph.com/-/debug/grafana/explore?orgId=1&left=%5B%221644239097280%22,%221644260697280%22,%22Prometheus%22,%7B%22expr%22:%22cadvisor_container_cpu_usage_percentage_total%7Bname%3D~%5C%22%5Erepo-updater.*%5C%22%7D%22,%22interval%22:%22100ms%22,%22datasource%22:%22Prometheus%22,%22exemplar%22:true,%22requestId%22:%22Q-5613d068-0719-4e42-a073-852f2e95c8cc-0A%22%7D%5D)

daxmc99:
I dived deeper into `cpu_shares` vs `cpus`. Under the hood, this is equivalent Kubernetes Requests and Limits. 
Essentially, this PR would be removing the `limits` and replacing them with requests. 

This reduces the burden to manually increase or change the CPUs value. However, our monitoring today uses [this](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@main/-/blob/docker-images/prometheus/config/cadvisor_rules.yml?L8) to determine when an instance is nearing its limits.

daxmc99:
Closing for https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/760 & https://github.com/sourcegraph/customer/issues/700 

This fix looks like it would break our  grafana charts and wouldn't give us the value we hope for. 
Let's settle for adding alerting for now

PR #757, opened 2022-02-09
log all errors on exit

davejrt:
catch all errors on exit, rather than in the testing function. Ensures that errors are logged during any failure point during the script.


danieldides (approved):
🙏

davejrt:
closes https://github.com/sourcegraph/sourcegraph/issues/30861

PR #760, opened 2022-02-14
Remove GOMAXPROC env var

### LINKED ISSUE #30214: Investigate upgrading docker-compose on managed instances
Use `cpu_shares` in our Docker-Compose deployments for managed instances instead of `cpus` to reduce the operational overhead in scaling services on our managed instances according to core count.

----
**old info / context**

Is this something we should plan on doing? Just for managed instances? To customer deployments as well?

https://sourcegraph.slack.com/archives/C030K5GRW7K/p1643152290009900?thread_ts=1643152162.009800&cid=C030K5GRW7K

----

_This action was created from [Incident 75 (Managed instance alerts not received)](https://app.incident.io/incidents/75?ctx=tqLXrxAp6JbXD7q3OxCRqBV2D%2B3w6iNBWNwfbtvXgQUZT1%2Fz5AwBWFj%2Bdcvj7ySu4klA%2BMzsWIvMHWdOF7qq13Txn%2Bzpn5Ouap0xkEdSAUTQlj6BaXwB1OggE7qi8LiB5aRU7aodS5jjXMU%3D) by Kevin Wojkovich, using [incident.io](https://app.incident.io) 🔥_

daxmc99:
Closes https://github.com/sourcegraph/sourcegraph/issues/30214 

We use [automaxproc](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24%40main+automaxprocs&patternType=literal) in our codebase already. This env var just adds more docker-specific config that needs to be set. 

### Checklist

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -74,7 +74,6 @@ services:
     mem_limit: '8g'
     environment:
       - DEPLOY_TYPE=docker-compose
-      - GOMAXPROCS=12
       - JAEGER_AGENT_HOST=jaeger
       - PGHOST=pgsql
       - CODEINTEL_PGHOST=codeintel-db
@@ -118,7 +117,6 @@ services:
     mem_limit: '8g'
     environment:
       - DEPLOY_TYPE=docker-compose
-      - GOMAXPROCS=4
       - PGHOST=pgsql
       - CODEINTEL_PGHOST=codeintel-db
       - CODEINSIGHTS_PGDATASOURCE=postgres://postgres:password@codeinsights-db:5432/postgres
@@ -150,7 +148,6 @@ services:
       codeintel-db:
         condition: service_healthy
 
-
   # Description: Stores clones of repositories to perform Git operations.
   #
   # Disk: 200GB / persistent SSD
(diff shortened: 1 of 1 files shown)


docker-compose/docker-compose.yaml
```diff
@@ -547,14 +536,14 @@ services:
   # would be bad but it can be rebuilt given enough time.)
   codeinsights-db:
     container_name: codeinsights-db
-    image: "index.docker.io/sourcegraph/codeinsights-db:insiders@sha256:72474269adad3a974c59323e063b45bb5933ec8121d5a34119bef504d7854f36"
+    image: 'index.docker.io/sourcegraph/codeinsights-db:insiders@sha256:72474269adad3a974c59323e063b45bb5933ec8121d5a34119bef504d7854f36'
```
caugustus-sourcegraph, line 539: Thank you!!

caugustus-sourcegraph (approved):
Is there any testing we need to do here to confirm automaxproc is working? Or do we already pretty much know that from the Kubernetes deployments?

daxmc99:
It is hard to directly test but since our metrics leverage `cpu_quota` and so does this package (https://sourcegraph.com/github.com/uber-go/automaxprocs/-/blob/internal/cgroups/cgroups.go?L40:2) if our container metrics stop reporting then we can infer this is working.

PR #761, opened 2022-02-16
Add migrator to pure-docker deployment

caugustus-sourcegraph:
Closes https://github.com/sourcegraph/sourcegraph/issues/31256

### Checklist

### DIFF
--- a/pure-docker/deploy-migrator.sh
+++ b/pure-docker/deploy-migrator.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+set -e
+
+# Description: Performs database migrations
+#
+# Network: 100mbps
+# Liveness probe: n/a
+# Ports exposed to other Sourcegraph services: none
+# Ports exposed to the public internet: none
+#
+
+docker run --detach \
+    --name=migrator \
+    --network=sourcegraph \
+    --restart=on-failure \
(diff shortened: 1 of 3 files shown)


michaellzc (approved):
lgtm

Should we highlight the migrator change in the pure_docker release notes for 3.37.0?

If we do, maybe we can start a thread in #release-guild so the next captain won't forget.

caugustus-sourcegraph:
> Should we highlight the migrator change in the pure_docker release notes for 3.37.0?

Yes, all the upgrade docs will need some documentation about the change - it's a pretty big one. I think Eric was working on a blog post that we'll be able to reference. I don't think the release captain should be responsible for writing the doc updates though; it can be done ahead of time. There's a separate conversation happening about what comms are needed / who's in charge.

michaellzc:
> > Should we highlight the migrator change in the pure_docker release notes for 3.37.0?
> 
> Yes, all the upgrade docs will need some documentation about the change - it's a pretty big one. I think Eric was working on a blog post that we'll be able to reference. I don't think the release captain should be responsible for writing the doc updates though; it can be done ahead of time. There's a separate conversation happening about what comms are needed / who's in charge.

yeah, as long as we reference the blog post from the release notes somewhere should be good enough 👍

PR #775, opened 2022-03-07
Reduce default resources on migrator container

caugustus-sourcegraph:
Adjust the defaults on the migrator container to be consistent with https://github.com/sourcegraph/deploy-sourcegraph/pull/4095. These are still probably too generous, but it's closer!

### Checklist

* [X] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph/pull/4095) change:

### Test plan

Ran docker-compose locally and confirmed migrator still worked.


danieldides (approved):
Realistically these can be lowered to basically nothing. The migrator just issues the SQL, the DB is doing all the work. Thank you for hunting down these values.

PR #776, opened 2022-03-08
Migrate codeinsights-db to vanilla postgres image

caugustus-sourcegraph:
Switch codeinsights to use the vanilla Postgres image.

Due to the uid mismatch between TimescaleDB and our Postgres image, we need to change existing file ownership (uid 70) to match the new uid (999). In this approach, we're running `chown` as a standalone container that will run before codeinsights-db starts up. This container can be removed in the 3.39 release.

Three concerns:
- `service_completed_successfully` requires a newer version of Docker Compose than we currently require (https://github.com/sourcegraph/sourcegraph/issues/32077). If this is a dealbreaker we may be able to get away with just letting codeinsights-db crash until the chown runs. **Update**: I think this is ok, we can make an update in the upgrade doc to tell customers to run the command separately.
- The performance cost of running the `chown` command is unknown. This needs to be tested against a larger dataset (insightstest), although the number of files involved should not be too unbearable. If it's too bad, we can always fall back to requiring customers to run a manual migration step. **Update:** no issue seen against insightstest
- The `sed` command in the chown container is a tiny bit risky. It definitely works for the pre-built postgresql.conf but if a customer has added their own things get messy. We'll need to rely on adding a warning during the upgrade notes (although I think there are very few customers who would have customized their conf in docker-compose).

### Checklist

### Test plan

Run through a migration on insightstest and measure performance / UX.

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -48,6 +48,8 @@ services:
     depends_on:
       pgsql:
         condition: service_healthy
+      codeinsights-db:
+        condition: service_healthy
       codeintel-db:
         condition: service_healthy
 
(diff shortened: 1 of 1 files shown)


docker-compose/docker-compose.yaml
```diff
@@ -619,31 +621,83 @@ services:
...
+  #
+  codeinsights-chown:
+    container_name: codeinsights-chown
+    image: index.docker.io/sourcegraph/alpine-3.12:insiders@sha256:290c7a045f18b8dab3021e6d5667f751b44c5bd488c408ebd387a802a4bf4876
+    cpus: 0.1
+    mem_limit: '10m'
+    command:
+      - "sh"
+      - "-c"
+      - if [ -d /var/lib/postgresql/data/ ]; then chown -R 999:999 /var/lib/postgresql/data/; sed -r -i "s/[#]*\s*(shared_preload_libraries)\s*=\s*'timescaledb(.*)\'/\1 = '\2'/;s/,'/'/" /var/lib/postgresql/data/pgdata/postgresql.conf; fi
```
caugustus-sourcegraph, line 642: The sed command fixes https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/777 and is just a reversal of the change run in the timescaledb image (https://github.com/timescale/timescaledb-docker/blob/master/Dockerfile#L79).

docker-compose/docker-compose.yaml
```diff
@@ -619,31 +621,83 @@ services:
...
   # Ports exposed to the public internet: none
   #
-  # Note: You should deploy this as a container, do not try to connect it to your external
-  # Postgres deployment (TimescaleDB is a bit special and most hosted Postgres deployments
-  # do not support TimescaleDB, the data here is akin to gitserver's data, where losing it
-  # would be bad but it can be rebuilt given enough time.)
   codeinsights-db:
     container_name: codeinsights-db
-    image: 'index.docker.io/sourcegraph/codeinsights-db:insiders@sha256:82fe7f91dbdca1cd397a80f1023d9da4e9c6af103bc36866d800e3a8279eea00'
+    image: 'index.docker.io/sourcegraph/postgres-12-alpine:135107_2022-03-03_9498a8bd3366@sha256:e26b159dc7c0c47d136886390c899816e669a3c2c1ead689bdad0b610364e45e'
```
caugustus-sourcegraph, line 658: Thoughts on whether we continue using a custom DB image? We could reuse the standard postgres-12-alpine image (less maintenance for customers running a private registry). Or we can keep building a custom `codeinsights-db` image that wraps our postgres image.

michaellzc: I like using the our standard postgres-12-alpine image. there's no good reason unless we are going to add custom config on top of the `soucegraph/postgres-12-alpine` image.

michaellzc:
> The performance cost of running the chown command is unknown. This needs to be tested against a larger dataset (insightstest), although the number of files involved should not be too unbearable. If it's too bad, we can always fall back to requiring customers to run a manual migration step.

Like you said, `chown` is only a problem when there're a lot of files, but pg shouldn't have a lot. 

> `service_completed_successfully` requires a newer version of Docker Compose than we currently require

what's the behaviour of running `docker-compose up -d` with an older version of docker compose? Will it result some weird race condition that codeinsights-db starts up before the chown container is finished? Or the docker-compose will just fail to start because it contains unsupported feature.

Additionally, taking `The sed command in the chown container is a tiny bit risky.` into account, I am thinking

1. add a sanity check to the script. If the conf file has been customized, error out immediately and point users to a manual migration guide.

2.1 If docker-compose will fail to run on older version, we can probably get away of utilizing `service_completed_successfully` feature.

2.2 If it doesn't and it caused weird behaviour, maybe we should make the migration an explicit manual step, like an extra `docker-compose.migration.yaml` file.

caugustus-sourcegraph:
A few updates after testing:

If the docker-compose version is too old, docker-compose up fails with the following error:
```The Compose file './docker-compose.yaml' is invalid because:
services.codeinsights-db.depends_on.codeinsights-chown.condition contains "service_completed_successfully", which is an invalid type, it should be a service_started, or a service_healthy
```
Running workloads are unaffected, so we should be ok with just noting this in the upgrade docs.

Running the chown against the insightstest data set took a negligible amount of time. The database does run through a [reindexing procedure](https://github.com/sourcegraph/sourcegraph/blob/main/docker-images/postgres-12-alpine/rootfs/reindex.sh), which probably isn't required, but is safer to keep for future-proofing.

caugustus-sourcegraph:
Closing this PR since I've changed the approach pretty significantly, will open a new one

PR #777, opened 2022-03-08
Swap codeinsights to vanilla postgres

caugustus-sourcegraph:
Just demoing the swap I'm testing out. `postgres:12.7-alpine` has the same uid as timescale so there are no permissions issues.

Error seen when trying to swap an existing deployment:
```
2022-03-08 21:00:36.119 UTC [1] FATAL:  could not access file "timescaledb": No such file or directory
2022-03-08 21:00:36.119 UTC [1] LOG:  database system is shut down

PostgreSQL Database directory appears to contain a database; Skipping initialization
```

Steps to reproduce:
- Start up docker-compose deployment running v3.37.0
- Update timescaleDB image reference to `postgres:12.7-alpine`
- Error!

### Checklist

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -632,7 +632,7 @@ services:
   # would be bad but it can be rebuilt given enough time.)
   codeinsights-db:
     container_name: codeinsights-db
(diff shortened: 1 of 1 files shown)


caugustus-sourcegraph:
Problem resolved - the issue was that the timescale extension was added to the shared_preload_libraries postgres config.

PR #782, opened 2022-03-15
Frontend containers now wait for the migrator to complete

caugustus-sourcegraph:
The frontend-internal container now depends on the migrator completing successfully, preventing an outage when the migrator fails and fixing some flaky tests / startup failures.

This change was originally part of the initial migrator deployment but had to be rolled back in order for us to upgrade the minimum version of docker-compose (done in https://github.com/sourcegraph/sourcegraph/pull/32631).

### Checklist

### Test plan

New deployment started successfully.
Upgrade of deployment was successful.
A failed migrator container blocked the frontend containers from rolling out.

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -198,6 +198,8 @@ services:
         condition: service_healthy
       codeintel-db:
         condition: service_healthy
+      migrator:
+        condition: service_completed_successfully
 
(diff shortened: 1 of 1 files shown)


michaellzc (changes requested):
🎉

michaellzc (approved):
wrong button

PR #791, opened 2022-04-05
Switch the Checkov step to soft-fail

ferozsalam:
Until we can move to using a pinned Checkov image version – the lack of image
pinning is causing the build step to break unpredictably.

### Checklist

* [x] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: https://github.com/sourcegraph/deploy-sourcegraph/pull/4106
* N/A - All images have a valid tag and SHA256 sum

### Test plan

Confirm that the build step no longer hard fails

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -34,3 +34,4 @@ steps:
   - command: .buildkite/ci-checkov.sh
     label: ":lock: security - checkov"
     agents: { queue: "standard" }
+    soft_fail: true


andreeleuterio (approved):

caugustus-sourcegraph (approved):

ferozsalam:
@caugustus-sourcegraph are you able to merge this change in please? I don't have the necessary permissions myself.

PR #792, opened 2022-04-05
Add support to customize pg conf

michaellzc:
Part of
- https://github.com/sourcegraph/sourcegraph/issues/32918
- https://github.com/sourcegraph/sourcegraph/issues/33016

Learn more https://docs.google.com/document/d/1MzTpztUW-R8qARjQjYngA62LM64Y6OSImI04J_DPXTQ/edit 

### Checklist

### Test plan

```sh
cat docker-compose.yaml | yq '.services[].cpus = 0' > docker-compose.local-dev.yaml
```

```sh
docker-compose -f docker-compose.local-dev.yaml up -d
```


codeinsights-db/conf/postgresql.conf
```diff
@@ -0,0 +1,781 @@
...
+shared_preload_libraries = ''
+max_locks_per_transaction = 64
+
+
+#------------------------------------------------------------------------------
+# CUSTOMIZED OPTIONS
+#------------------------------------------------------------------------------
+
+# Add you customization here
+# Learn more: https://docs.sourcegraph.com/admin/config/postgres-conf
```
michaellzc, line 781: anything above is just a clone of the stock postgresql.conf

codeintel-db/conf/postgresql.conf
```diff
@@ -0,0 +1,779 @@
...
+random_page_cost = 1.1
+effective_cache_size = 3GB
+
+
+#------------------------------------------------------------------------------
+# CUSTOMIZED OPTIONS
+#------------------------------------------------------------------------------
+
+# Add you customization here
+# Learn more: https://docs.sourcegraph.com/admin/config/postgres-conf
```
michaellzc, line 779: anything above is just a clone of the stock postgresql.conf

pgsql/conf/postgresql.conf
```diff
@@ -0,0 +1,779 @@
...
+max_wal_size = 8GB
+min_wal_size = 2GB
+random_page_cost = 1.1
+effective_cache_size = 3GB
+
+
+#------------------------------------------------------------------------------
+# CUSTOMIZED OPTIONS
+#------------------------------------------------------------------------------
+
```
michaellzc, line 777: anything above is just a clone of the stock postgresql.conf

pgsq.diff
```diff
@@ -0,0 +1,875 @@
+3a4
```
caugustus-sourcegraph, line 1: Looks like this was checked in by mistake

codeinsights-db/conf/postgresql.conf
```diff
@@ -0,0 +1,781 @@
...
+autovacuum_naptime = 10
+shared_preload_libraries = ''
+max_locks_per_transaction = 64
+
+
+#------------------------------------------------------------------------------
+# CUSTOMIZED OPTIONS
+#------------------------------------------------------------------------------
+
+# Add you customization here
```
caugustus-sourcegraph, line 780:
```suggestion
# Add your customization here
```
(same for the other conf's)

codeinsights-db/conf/postgresql.conf
```diff
@@ -0,0 +1,781 @@
...
+#include_if_exists = '...'		# include file only if it exists
+#include = '...'			# include file
+
+
+#------------------------------------------------------------------------------
+# SOURCEGRAPH RECOMMENDED OPTIONS
+#------------------------------------------------------------------------------
+
+# Below is Sourcegraph recommended Postgres configuration.
+# Your should apply your own customization in CUSTOMIZED OPTIONS section below
```
caugustus-sourcegraph, line 750:
```suggestion
# You should apply your own customization in the CUSTOMIZED OPTIONS section below
```

codeinsights-db/conf/postgresql.conf
```diff
@@ -0,0 +1,781 @@
...
+# default postgresql.conf.  Note that these are directives, not variable
+# assignments, so they can usefully be given more than once.
+
+#include_dir = '...'			# include files ending in '.conf' from
+					# a directory, e.g., 'conf.d'
+#include_if_exists = '...'		# include file only if it exists
+#include = '...'			# include file
+
+
+#------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
michaellzc: > Do you think we could get rid of everything above our customizations? Those should all be the defaults and not required to be in this file (I think). That would make our changes much more visible for customers instead of being at the end of a 800-line file

It looks like we can't. The default `listen_address` is localhost (https://postgresqlco.nf/doc/en/param/listen_addresses/), the migrator (or any other services) will never able to reach pgsql.

The right way would have been do not touch the default postgresql.conf, but only insert a `include_dir '/conf'` at the end. By doing so, PG can load all conf files during runtime.

I am debating whether this is something we want to do, feels like a big change since we may need to change the postgres image https://github.com/sourcegraph/sourcegraph/tree/main/docker-images/postgres-12-alpine

thoughts?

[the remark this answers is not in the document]
caugustus-sourcegraph: Ugh, yeah. There's about 10 settings that are customized from the defaults hidden away in that file. I still think it would be beneficial to remove all the commented-out sections and only leave the things that are being explicitly set.

Your point about the `include_dir` is good, but it does feel like a pretty large task with some risk of losing customer overrides.

How do you feel about removing just the commented out sections?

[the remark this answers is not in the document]
caugustus-sourcegraph: Ugh, yeah. There's about 10 settings that are customized from the defaults hidden away in that file. I still think it would be beneficial to remove all the commented-out sections and only leave the things that are being explicitly set.

Your point about the `include_dir` is good, but it does feel like a pretty large task with some risk of losing customer overrides.

How do you feel about removing just the commented out sections?

codeinsights-db/conf/postgresql.conf
```diff
@@ -0,0 +1,91 @@
...
+#
+# This file consists of lines of the form:
+#
+#   name = value
+#
+# (The "=" is optional.)  Whitespace may be used.  Comments are introduced with
+# "#" anywhere on a line.  The complete list of parameter names and allowed
+# values can be found in the PostgreSQL documentation.
+#
+# The commented-out settings shown in this file represent the default values.
```
caugustus-sourcegraph, line 13: We can remove L13-L25 (maybe even L27-31 but eh).

Honestly I don't think we need to include any of the preamble - anyone who wants to configure Postgres will need to look up guidance anyway - but it's not a big deal.

codeinsights-db/conf/postgresql.conf
```diff
@@ -0,0 +1,91 @@
...
+#------------------------------------------------------------------------------
+# SOURCEGRAPH RECOMMENDED OPTIONS
+#------------------------------------------------------------------------------
+
+# Below is Sourcegraph recommended Postgres configuration.
+# You should apply your own customization in the CUSTOMIZED OPTIONS section below
+# to avoid merge conflict in the future.
+#
+# NOTES: Options below is based on the default resource configuration of `codeinsights-db` container.
+# Depending on your environment, you may need to adjust the settings in CUSTOMIZED OPTIONS below
```
caugustus-sourcegraph, line 65: This note basically says the same thing as above - do we need both it and L61-62?

codeinsights-db/conf/postgresql.conf
```diff
@@ -0,0 +1,91 @@
...
+
+shared_buffers = 509546kB
+work_mem = 3184kB
+maintenance_work_mem = 254773kB
+effective_io_concurrency = 200
+max_worker_processes = 19
```
caugustus-sourcegraph, line 71: Oh boy - I started looking into this because I thought 19 was such a weird number, and it turns out this is the [initial configuration](https://github.com/sourcegraph/deploy-sourcegraph/pull/1851) from Feb 2021 and tuned for timescaledb:
```
The postgres.conf included here comes from TimescaleDB itself, I included comments in the ConfigMap about differences and how to use timescaledb-tune to generate a good config file.
```

So, now that we're using vanilla Postgres, these numbers might not be as relevant. I think it's ok for this PR to keep those defaults but we should ask the Code Insights team to review the settings soon.

caugustus-sourcegraph (approved):
👍  Some followup actions but overall it's a great improvement! We can copy the same streamlined conf to helm too. Maybe even kustomize? Not sure it's worth introducing huge merge conflicts there.

Although.. is there any reason we shouldn't build these defaults in to the docker images? I guess it could be a bad idea for local development.

codeinsights-db/conf/postgresql.conf
caugustus-sourcegraph: Asked Code Insights team to investigate here: https://sourcegraph.slack.com/archives/C014ZCKMCAV/p1649421612224539

codeinsights-db/conf/postgresql.conf
michaellzc, line 65: > This note basically says the same thing as above - do we need both it and L61-62?

@caugustus-sourcegraph 

In the second paragraph, I was trying to say, the configuration is based on the default `mem_limit` and `cpus`.

how about

```suggestion
# Below is Sourcegraph recommended Postgres configuration based on the default resource configuration.
# You should apply your own customization in the CUSTOMIZED OPTIONS section below
# to avoid merge conflict in the future.
```

caugustus-sourcegraph: Looks great!

caugustus-sourcegraph (approved):

PR #798, opened 2022-04-11
Update prometheus_targets.yml for codeinsights-db

michaellzc:
Need to target the exporter container

### Checklist


caugustus-sourcegraph (approved):
TY!

PR #802, opened 2022-04-20
release: sourcegraph@3.39.0

keegancsmith:
This pull request is part of the Sourcegraph 3.39.0 release.

* [Release batch change](https://k8s.sgdev.org/organizations/sourcegraph/batch-changes/release-sourcegraph-3.39.0)
* [Tracking issue](https://github.com/sourcegraph/sourcegraph/issues/32836)

### :warning: Additional changes required

These steps must be completed before this PR can be merged, unless otherwise stated. Push any required changes directly to this PR branch.

- [x] Follow the [release guide](https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/RELEASING.md) to complete this PR 

cc @keegancsmith

### Test plan

CI checks in this repository should pass, and a manual review should confirm if the generated changes are correct.

### DIFF
--- a/docker-compose/db-only-migrate.docker-compose.yaml
+++ b/docker-compose/db-only-migrate.docker-compose.yaml
@@ -12,7 +12,7 @@ services:
   #
   pgsql:
     container_name: pgsql
-    image: 'index.docker.io/sourcegraph/postgres-12-alpine:135107_2022-03-03_9498a8bd3366@sha256:e26b159dc7c0c47d136886390c899816e669a3c2c1ead689bdad0b610364e45e'
+    image: 'index.docker.io/sourcegraph/postgres-12-alpine:3.39.0@sha256:ab88bfe18a8cded3f1ac653ab9b314b4a22a1b96fef5b9bff462ece2ff912559'
     cpus: 4
     mem_limit: '2g'
     healthcheck:
@@ -29,7 +29,7 @@ services:
 
   codeintel-db:
     container_name: codeintel-db
-    image: 'index.docker.io/sourcegraph/codeintel-db:insiders@sha256:cd901b1180229101cc8c6deebda24c14f6224295cd2153cafe0665199499c342'
+    image: 'index.docker.io/sourcegraph/codeintel-db:3.39.0@sha256:1a6038e7f6fa58295b208a1034f6370a9889f571ba41d905a9bebc1d5601cc40'
     cpus: 4
     mem_limit: '2g'
     healthcheck:
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -13,7 +13,7 @@ services:
   # for this container will need to be updated to reflect the new connection information.
   migrator:
     container_name: migrator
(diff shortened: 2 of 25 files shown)


keegancsmith:
> Smoke test: ensure Docker Compose upgrades work
> ⚠️ This test now runs in Buildktie, in the qa pipeline under the Sourcegraph Upgrade step, you can validate [the results of the CI run](https://buildkite.com/sourcegraph)

This pipeline doesn't seem to have had tests in a long time. Is this still accurate? cc @caugustus-sourcegraph @camdencheek

caugustus-sourcegraph:
@keegancsmith - Ugh, no, that's not accurate at all. Ignore everything about that page except what's listed in https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/RELEASING.md#releasing-pure-docker.

The CI tests you'll want to check are automatically run on each PR and block you from merging until they're successful (this pipeline https://buildkite.com/sourcegraph/deploy-sourcegraph-docker).

keegancsmith:
@camdencheek I noticed in your deploy-sourcegraph-docker change you didn't include `prometheus/prometheus_targets.yml` changes. But I think those are config files that are part of the upgrade? I'm gonna include them in mine, can adjust.

camdencheek:
That sounds correct. That sounds like an oversight on my part.

caugustus-sourcegraph:
@keegancsmith - I think mentioning [this line](https://github.com/sourcegraph/deploy-sourcegraph-docker/commit/8cbf2865535f67a108d4091db19a3b733ca24a71#diff-916162e35509bb582798c4306953fec9f43779d82420cb4435576e2873869f78R17) in particular is worthwhile, since it edits a config file in-place and customers might prefer to do that themselves. The intent is they need to remove timescaledb from the `shared_preload_libraries` of their codeinsights postgres config.

We probably shouldn't include the prometheus changes at all - they're updating the targets to point to some database exporters that don't run in pure-docker, so that will never work. I don't think it matters too much though, the customer we know uses this has their own monitoring setup.

keegancsmith:
> @keegancsmith - I think mentioning [this line](https://github.com/sourcegraph/deploy-sourcegraph-docker/commit/8cbf2865535f67a108d4091db19a3b733ca24a71#diff-916162e35509bb582798c4306953fec9f43779d82420cb4435576e2873869f78R17) in particular is worthwhile, since it edits a config file in-place and customers might prefer to do that themselves. The intent is they need to remove timescaledb from the `shared_preload_libraries` of their codeinsights postgres config.

@caugustus-sourcegraph will do

> We probably shouldn't include the prometheus changes at all - they're updating the targets to point to some database exporters that don't run in pure-docker, so that will never work. I don't think it matters too much though, the customer we know uses this has their own monitoring setup.

aah I didn't realise prometheus wasn't included. I'm happy to undo that change. There is a fun branch protection rule preventing me from force pushing to the customer replica branch, luckily I am an admin and can override it for now.

caugustus-sourcegraph (approved):

caugustus-sourcegraph:
@keegancsmith Nah, don't worry about it - the prometheus config was already broken, and I doubt it's really used directly. More likely it's just a reference for the customer so they know to check their own configuration.

PR #804, opened 2022-04-23
Fix cadvisor unable to report oomkill events

michaellzc:
context: https://sourcegraph.slack.com/archives/C02Q9G9A59S/p1650667926243749?thread_ts=1650662753.412789&cid=C02Q9G9A59S

on managed instances, all cadvisor is reporting such an error
```sh
$ docker logs cadvisor 2>&1  | grep "OOM"
W0420 16:11:04.614936       1 manager.go:296] Could not configure a source for OOM detection, disabling OOM events: open /dev/kmsg: no such file or directory
```

We think this is the cause of `container_oom_events_total` is always zero.

### Checklist

### Test plan

Pending for validation https://github.com/sourcegraph/deploy-sourcegraph-managed/pull/444

it works

CleanShot 2022-04-25 at 17 39 58

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -491,6 +491,11 @@ services:
     image: 'index.docker.io/sourcegraph/cadvisor:insiders@sha256:3544dba5cd99c429ac62f4773454c8406abe126ba340f34e457b38d828c63395'
     cpus: 1
     mem_limit: '1g'
+    # You may set `privileged` to `false and `cadvisor` will run with reduced privileges.
+    # `cadvisor` requires root privileges in order to display provisioning metrics.
+    # These metrics provide critical information to help you scale the Sourcegraph deployment.
+    # If you would like to bring your own infrastructure monitoring & alerting solution,
+    # you may want to remove the `cadvisor` container completely 
     privileged: true
     volumes:
       - '/:/rootfs:ro'
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -491,6 +491,7 @@ services:
     image: 'index.docker.io/sourcegraph/cadvisor:insiders@sha256:3544dba5cd99c429ac62f4773454c8406abe126ba340f34e457b38d828c63395'
     cpus: 1
     mem_limit: '1g'
+    privileged: true
     volumes:
       - '/:/rootfs:ro'
       - '/var/run:/var/run:ro'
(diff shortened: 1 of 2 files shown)


michaellzc:
Current dependencies on/for this PR:
* master
  * **PR #804** <a href="https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-docker/804" target="_blank">Graphite</a>  (this PR)

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/deploy-sourcegraph-docker/804?utm_source=stack-comment).

docker-compose/docker-compose.yaml
```diff
@@ -499,6 +500,8 @@ services:
       - '/dev/disk/:/dev/disk:ro'
       # Uncomment to enable container monitoring on MacOS
       # - '/var/run/docker.sock:/var/run/docker.sock:ro'
+    devices:
+      - '/dev/kmsg'
```
michaellzc, line 504: all flags are documented here https://github.com/google/cadvisor/blob/master/README.md#quick-start-running-cadvisor-in-a-docker-container

these two flags were not in the readme when MI was invented in the first place (I think) and that's how we missed them

https://github.com/google/cadvisor/issues/2150
https://github.com/google/cadvisor/pull/2545

docker-compose/docker-compose.yaml
```diff
@@ -491,6 +491,7 @@ services:
     image: 'index.docker.io/sourcegraph/cadvisor:insiders@sha256:3544dba5cd99c429ac62f4773454c8406abe126ba340f34e457b38d828c63395'
     cpus: 1
     mem_limit: '1g'
+    privileged: true
```
michaellzc, line 494: I suppose some users would be worried with privileged (root) container, anything we can do to address future concern?

caugustus-sourcegraph: Maybe add a comment saying it can be set to false, in case anyone has concerns? More likely they'd just disable cadvisor entirely but it can't hurt.

caugustus-sourcegraph (approved):
Nice! Does deploy-sourcegraph need a similar change? Or pure-docker?

docker-compose/docker-compose.yaml
michaellzc, line 494:
```suggestion
    # You may set `privileged` to `false and `cadvisor` will run with reduced privileges.
    # `cadvisor` requires root privileges in order to display provisioning metrics.
    # These metrics provide critical information to help you scale the Sourcegraph deployment.
    # If you would like to bring your own infrastructure monitoring & alerting solution,
    # you may want to remove the `cadvisor` container completely 
    privileged: true
```

caugustus-sourcegraph (approved):

michaellzc:
> Nice! Does deploy-sourcegraph need a similar change? Or pure-docker?

Just added the changes for pure-docker as well

for k8s, we should be able to get it working https://stackoverflow.com/a/59291859, but it will wait until later this week or next week.

daxmc99 (approved):

PR #805, opened 2022-04-25
Clarify db-only-migrate.docker-compose.yaml

Kelvin-Lee:
Wondering if this very, very small change is more correct. There are technically now two services in this file, right? And they're both postgres? If the change is more correct, then I will also create the same change for our Kubernetes repo.

### Checklist

* [x] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: Not applicable.
### Test plan
None. Though confidence gained from [Slack thread](https://sourcegraph.slack.com/archives/C02E4HE42BX/p1650918019969739), between myself and @kevinwojo .

### DIFF
--- a/docker-compose/db-only-migrate.docker-compose.yaml
+++ b/docker-compose/db-only-migrate.docker-compose.yaml
@@ -1,4 +1,4 @@
-# This file contains only the pgsql service definition, and is used during
+# This file contains the pgsql service definitions, and is used during
 # the sourcegraph/server -> docker-compose migration process.
 #
 # 🚨This file MUST be kept in sync with the pgsql definition in docker-compose/docker-compose.yaml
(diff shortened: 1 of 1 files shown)


docker-compose/db-only-migrate.docker-compose.yaml
```diff
@@ -1,4 +1,4 @@
-# This file contains only the pgsql service definition, and is used during
+# This file contains the pgsql service definitions, and is used during
```
caugustus-sourcegraph, line 1: Maybe database instead of pgsql? _Technically_ pgsql refers specifically to the main SG database deployment, and codeintel-db is a separate postgres deployment.

Kelvin-Lee:
Hey @caugustus-sourcegraph , looks like this need a formal review. Kevin and I chatted a bit [here](https://sourcegraph.slack.com/archives/C02E4HE42BX/p1650918796181309?thread_ts=1650918019.969739&cid=C02E4HE42BX), for extra context. I think it's all good to go, just a tiny change.

caugustus-sourcegraph (approved):
👍  Thanks for the fix!

Kelvin-Lee:
Made a new commit, @caugustus-sourcegraph 😅 Let me know if we should:

(a) Go back to the original fix (just changing "definition" -> "definitions")
(b) This new commit is fine.
(c) If you want to write the commit instead with your expert wording 😄

PR #807, opened 2022-04-26
replica: add postgres configuration files into long-running branch

kevinwojo:
It appears that 3.39 did not include postgres configuration files in the "(changes unrelated to upgrade)" commit.

```
On branch 3.39-customer-replica
Your branch is up to date with 'origin/3.39-customer-replica'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	codeinsights-db/
	codeintel-db/
	pgsql/
```

Context: https://github.com/sourcegraph/customer/issues/854#issuecomment-1109916993

Since new customer-replica branches are based on the previous version's branch, this will be a fix-forward.

### Checklist

* [x] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change:
### Test plan
Manually tested, this branch is also more for documentation.

### DIFF
--- a/codeinsights-db/conf/postgresql.conf
+++ b/codeinsights-db/conf/postgresql.conf
@@ -1,0 +1,57 @@
+#------------------------------------------------------------------------------
+# POSTGRESQL DEFAULT CONFIGURATION
+#------------------------------------------------------------------------------
+
+# Below is PostgreSQL default configuration.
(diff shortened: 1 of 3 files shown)


caugustus-sourcegraph (commented):
This PR seems incomplete - doesn't it also require the changes made to the docker volume mounts?

PR #809, opened 2022-04-27
3.39.1 customer replica

coury-clark:
### Checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,2 @@
 .github/
 *.md
-rootfs/etc/docker/daemon.json
--- a/prometheus/prometheus_targets.yml
+++ b/prometheus/prometheus_targets.yml
@@ -56,7 +56,7 @@
 - labels:
     job: codeinsights-db
   targets:
-    - codeinsights-db:9187
+    - codeinsights-db-exporter:9187
 - labels:
     job: redis-cache
   targets:
--- a/pure-docker/deploy.sh
+++ b/pure-docker/deploy.sh
@@ -10,7 +10,6 @@ source ./replicas.sh
 
 docker network create sourcegraph &> /dev/null || true
 
-./deploy-apache.sh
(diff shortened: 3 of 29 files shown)


.prettierignore
```diff
@@ -1,3 +1,2 @@
 .github/
 *.md
-rootfs/etc/docker/daemon.json
```
camdencheek, line 3: I don't think this is supposed to go away

prometheus/prometheus_targets.yml
```diff
@@ -56,7 +56,7 @@
 - labels:
     job: codeinsights-db
   targets:
-    - codeinsights-db:9187
+    - codeinsights-db-exporter:9187
```
camdencheek, line 59: Just double checking that this got manual eyes

pure-docker/deploy.sh
```diff
@@ -49,6 +48,5 @@ do
 done
 
 for i in $(seq 0 $(($NUM_FRONTEND - 1))); do ./deploy-frontend.sh $i; done
-# Not used in customer-replica branch.
-#./deploy-caddy.sh
+./deploy-caddy.sh
```
camdencheek, line 51: I think the changes in this file need to not be included as well

camdencheek (changes requested):

prometheus/prometheus_targets.yml
coury-clark: Yeah I don't really know what this is, but I don't think it's coming from the release tool

pure-docker/deploy.sh
coury-clark: Ah yikes - thanks!

prometheus/prometheus_targets.yml
camdencheek: Cool. Seems to be in the right commit either way.

PR #816, opened 2022-05-10
Remove container_name field

daxmc99:
This field prevents docker from automatically appending service names when scaling, (ie: 'worker.1','worker.2')

### Checklist

### Test plan

Tested locally with scale set

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -12,12 +12,10 @@ services:
   # If you're using a DB instance hosted outside of docker-compose, the environment variables
   # for this container will need to be updated to reflect the new connection information.
   migrator:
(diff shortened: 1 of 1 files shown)


danieldides:
I think this is a good change for managed instances but we need to make sure that it's well documented as this will be a non-trivial change for customers if they've manually scaled (by adding additional named containers) they will need to migrate their configs over to having a `deploy.replicas` field. 

We also need to make sure that there's a way to migrate any sort of persistent disks from the manually created containers to the newly auto-scaled containers.

daxmc99:
Proceeding to not due this based on comments in issue. Horizontal scaling of docker-compose components is going to be accomplished via CLI + docker-compose override files

PR #824, opened 2022-06-01
Make the Postgres healthchecks more lenient

chrismwendt:
See https://github.com/sourcegraph/deploy-sourcegraph/pull/4136

### Checklist

* [ ] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: https://github.com/sourcegraph/deploy-sourcegraph/pull/4136

### Test plan

N/A

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -659,7 +659,7 @@ services:
       interval: 10s
       timeout: 1s
       retries: 10
-      start_period: 15s
+      start_period: 1h
     volumes:
       - 'codeinsights-db:/var/lib/postgresql/data/'
       - '../codeinsights-db/conf:/conf/'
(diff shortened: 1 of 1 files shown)


docker-compose/docker-compose.yaml
```diff
@@ -559,7 +559,7 @@ services:
       test: '/liveness.sh'
       interval: 10s
       timeout: 1s
-      retries: 10
+      retries: 360
```
caugustus-sourcegraph, line 562: If the health of a container can fail for a full hour, is this healthcheck really even doing anything valuable? Perhaps modifying the start_period is a better option here, if this is intended to fix the same slow startup issue as in https://github.com/sourcegraph/deploy-sourcegraph/pull/4136.

https://docs.docker.com/engine/reference/builder/#healthcheck
> start period provides initialization time for containers that need time to bootstrap. Probe failure during that period will not be counted towards the maximum number of retries. However, if a health check succeeds during the start period, the container is considered started and all consecutive failures will be counted towards the maximum number of retries.

docker-compose/docker-compose.yaml
```diff
@@ -606,7 +606,7 @@ services:
...
```
caugustus-sourcegraph, line 609: Is it intentional that codeinsights-db isn't included in this change?

PR #828, opened 2022-06-14
Upgrade caddy

andreeleuterio:
This patches CVE-2021-42377 and CVE-2020-28928.
### Checklist

### Test plan
* CI checks
* @kevinwojo: I created a GCP compute instance and installed SG with Caddy using HTTPS via LetsEncrypt. I then upgraded Caddy to this latest version. Aside from providing a notice to our customers regarding the new `trusted_proxies` setting, I think this is safe.

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -85,6 +85,9 @@ services:
     volumes:
       - 'caddy:/caddy-storage'
       #
+      # IMPORTANT: if a customer uses a reverse proxy in front of Caddy
+      # the configuration files below must be updated to include trusted_proxies
+      # 
       # Comment out the following line when using HTTPS with either Let's Encrypt or custom certificates
       - '../caddy/builtins/http.Caddyfile:/etc/caddy/Caddyfile'
       #
@@ -103,10 +106,6 @@ services:
       #
       # Uncomment / update the following line when using HTTPS with custom certificates
       # - '/LOCAL/KEY/PATH.key:/sourcegraph.key'
-      # 
-      # Uncomment / update the following line if Caddy is behind a reverse proxy or load balancer
-      # IMPORTANT: the file below needs to be updated to include customer's ips
-      # - '../caddy/builtins/https.trusted-proxies.Caddyfile'
     ports:
       - '0.0.0.0:80:80'
       - '0.0.0.0:443:443'
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -103,6 +103,10 @@ services:
       #
       # Uncomment / update the following line when using HTTPS with custom certificates
       # - '/LOCAL/KEY/PATH.key:/sourcegraph.key'
+      # 
+      # Uncomment / update the following line if Caddy is behind a reverse proxy or load balancer
+      # IMPORTANT: the file below needs to be updated to include customer's ips
+      # - '../caddy/builtins/https.trusted-proxies.Caddyfile'
     ports:
       - '0.0.0.0:80:80'
       - '0.0.0.0:443:443'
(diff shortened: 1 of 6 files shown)


kevinwojo (commented):
We'll want to test this release a little more thoroughly.

From the [CHANGELOG](https://github.com/caddyserver/caddy/releases):

> ⚠️ Reverse proxy: Incoming X-Forwarded-* headers will no longer be automatically trusted, to prevent spoofing. Now, trusted_proxies must be configured to specify a list of downstream proxies which are trusted to have sent good values. You only need to configure trusted proxies if Caddy is not the first server being connected to. For example, if you have Cloudflare in front of Caddy, then you should configure this with Cloudflare's [list of IP ranges](https://www.cloudflare.com/en-ca/ips/).

We may have to provide some documentation guidance for our customers if they're in this position. 

I think we _could_ probably copy/paste this notice into the [Docker Compose upgrade notes](https://docs.sourcegraph.com/admin/updates/docker_compose#updating-a-docker-compose-sourcegraph-instance).

kevinwojo:
Accompanying release note: https://github.com/sourcegraph/sourcegraph/pull/37270

kevinwojo (approved):
Tested and added release note.

kevinwojo (changes requested):
Putting a soft block on this, we need to update a good chunk of docs before we can ship this.

I started typing up a courtesy message for CE and realized we have a lot to communicate.

andreeleuterio:
@kevinwojo I updated the example docker-compose file and added a sample file for `trusted_proxies` config.

docker-compose/docker-compose.yaml
```diff
@@ -103,6 +103,10 @@ services:
       #
       # Uncomment / update the following line when using HTTPS with custom certificates
       # - '/LOCAL/KEY/PATH.key:/sourcegraph.key'
+      # 
+      # Uncomment / update the following line if Caddy is behind a reverse proxy or load balancer
+      # IMPORTANT: the file below needs to be updated to include customer's ips
+      # - '../caddy/builtins/https.trusted-proxies.Caddyfile'
```
kevinwojo, line 109: What happens if they're already using one of the other http or https configuration files?

Perhaps we just add this commented-out block in the existing files?

andreeleuterio: I thought it bundled config files. My b, I shouldn't have assumed that. I'll clean this up later today.

andreeleuterio: Fixed!

kevinwojo:
This is on my list to test today. I am preparing a few internal examples to share with the team internally that we can use for validation of config syntax. 

Admittedly the Caddyfile syntax is a little nebulous ... documentation is hard to grok.

daxmc99:
Paired with @kevinwojo on this, determined that although we don't use the `X-Forwarded-For` header [much](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/sourcegraph%24+x-forwarded-for&patternType=literal) we might lose some analytics

I suggest we modify these caddy files to be 
```
# Routes all plain http requests to sourcegraph-frontend - suitable for local testing.
#
# Caddyfile documentation: https://caddyserver.com/docs/caddyfile

:80

# If the customer uses a reverse proxy in front of Caddy,
# add the reverse proxies IPs (or IP CIDR ranges) to the trusted_proxies list.
# More information in https://caddyserver.com/docs/caddyfile/directives/reverse_proxy
reverse_proxy {
to {$SRC_FRONTEND_ADDRESSES}
trusted_proxies 0.0.0.0/0
}
```

daxmc99:
@sourcegraph/security for approval of this approach

daxmc99:
Spoke with @andreeleuterio offline, and he approved this approach. I will commit and merge this.

kevinwojo (approved):

PR #838, opened 2022-07-14
docker-compose: add otel-collector override, dev override

bobheadxi:
Part of https://github.com/sourcegraph/sourcegraph/issues/38819 with @davejrt - this demonstrates how one might set up otel-collector in a docker-compose environment. I think getting it set up in s2 - the desired end state of https://github.com/sourcegraph/sourcegraph/issues/38819 - would just be applying the override and the new base manifest to the instance.

### Checklist

### Test plan

```sh
docker-compose \
-f docker-compose/docker-compose.yaml \
-f docker-compose/otel/docker-compose.yaml \
-f docker-compose/dev/docker-compose.yaml up
```

In site config:

```json
{
  "externalURL": "http://127.0.0.1:8080",
  "observability.tracing": {
    "type": "opentelemetry",
    "debug": true,
    "sampling": "all"
  }
}
```

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -13,7 +13,7 @@ services:
   # for this container will need to be updated to reflect the new connection information.
   migrator:
     container_name: migrator
-    image: 'index.docker.io/sourcegraph/migrator:insiders@sha256:e5a348fa3cbfca461a3a7490364bed90bdfe14de75b89e41acb2f13519c692bf'
+    image: 'index.docker.io/sourcegraph/migrator:insiders@sha256:82708de0e12667f95cde7ab9fb690e46277ce5c1691effe8adb7e268ac3d7c2f'
     cpus: 0.5
     mem_limit: '500m'
     command:
@@ -125,7 +125,7 @@ services:
   # service.
   sourcegraph-frontend-0:
     container_name: sourcegraph-frontend-0
-    image: 'index.docker.io/sourcegraph/frontend:insiders@sha256:ff468ff687e498aaa4d7dfbfb3a4232499e8f325c49a4e4bd7e1103fae26fc61'
+    image: 'index.docker.io/sourcegraph/frontend:insiders@sha256:5780439ac988443523ed4c7731dde912278bf2397060b765e6c25d7a9054660b'
     cpus: 4
     mem_limit: '8g'
     environment:
@@ -169,7 +169,7 @@ services:
   #
   sourcegraph-frontend-internal:
     container_name: sourcegraph-frontend-internal
-    image: 'index.docker.io/sourcegraph/frontend:insiders@sha256:ff468ff687e498aaa4d7dfbfb3a4232499e8f325c49a4e4bd7e1103fae26fc61'
+    image: 'index.docker.io/sourcegraph/frontend:insiders@sha256:5780439ac988443523ed4c7731dde912278bf2397060b765e6c25d7a9054660b'
     cpus: 4
     mem_limit: '8g'
     environment:
@@ -215,7 +215,7 @@ services:
   #
   gitserver-0:
     container_name: gitserver-0
-    image: 'index.docker.io/sourcegraph/gitserver:insiders@sha256:d0c399793afa1342e403f5fec37b526e93a90f5d285159e3b2847876c43fc78c'
+    image: 'index.docker.io/sourcegraph/gitserver:insiders@sha256:71df19add0eaea54ec80d4b162957fe241afd5bdd3ed0fe6b6a10c8e2b0737c4'
     cpus: 4
     mem_limit: '8g'
     environment:
@@ -284,7 +284,7 @@ services:
   #
   searcher-0:
     container_name: searcher-0
-    image: 'index.docker.io/sourcegraph/searcher:insiders@sha256:0bb91952077c1ca9afec16a781002f003036b90e40087d34ed0c8ed297efd5aa'
+    image: 'index.docker.io/sourcegraph/searcher:insiders@sha256:2e1d16b8f062539100e0eb0e3ec41016c72e8f7ccdba9a9e42efa8570016faed'
     cpus: 2
     mem_limit: '2g'
     environment:
@@ -311,7 +311,7 @@ services:
   #
   github-proxy:
     container_name: github-proxy
-    image: 'index.docker.io/sourcegraph/github-proxy:insiders@sha256:bbab6934d75f2d3779a72114f8d5ed12f5d89189e63748905cbac343bd668e93'
+    image: 'index.docker.io/sourcegraph/github-proxy:insiders@sha256:5c9fc95300ef3296ba8c1d87b7477ebd2b272c6c552d834ab7ec762695d351bb'
     cpus: 1
     mem_limit: '1g'
     environment:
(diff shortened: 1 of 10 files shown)


bobheadxi:
Struggling to get the collector to connect to jaeger:

```
2022-07-15T01:59:03.080Z        info    jaegerexporter@v0.54.0/exporter.go:186  State of the connection with the Jaeger Collector backend       {"kind": "exporter", "data_type": "traces", "name": "jaeger", "state": "CONNECTING"}
2022-07-15T01:59:03.098Z        warn    zapgrpc/zapgrpc.go:191  [core] [Channel #1 SubChannel #2] grpc: addrConn.createTransport failed to connect to {
  "Addr": "jaeger:14250",
  "ServerName": "jaeger:14250",
  "Attributes": null,
  "BalancerAttributes": null,
  "Type": 0,
  "Metadata": null
}. Err: connection error: desc = "transport: Error while dialing dial tcp 172.19.0.14:14250: connect: connection refused"       {"grpc_log": true}
2022-07-15T01:59:04.078Z        info    jaegerexporter@v0.54.0/exporter.go:186  State of the connection with the Jaeger Collector backend       {"kind": "exporter", "data_type": "traces", "name": "jaeger", "state": "TRANSIENT_FAILURE"}
```

docker-compose/otel/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
+version: '2.4'
+services:
+  otel-collector:
+    container_name: otel-collector
+    image: otel/opentelemetry-collector:0.54.0
```
bobheadxi, line 5: will follow up to switch this to a sourcegraph image as we get closer to making this generally available

michaellzc (commented):
nice, tested locally and it works.

questions:

- do we plan on shipping `otel-collector` to all customers as default? or will it remain as an opt-in feature?
- if it will remain as an opt-in feature, should we consider moving the env var override `OTEL_EXPORTER_OTLP_ENDPOINT=otel-collector:4317` into `otel/docker-compose.yaml` override file? Some customers are "too curious" about random env var that we have, this could avoid some confusion.

docker-compose/docker-compose.yaml
```diff
@@ -310,12 +315,13 @@ services:
...
   github-proxy:
     container_name: github-proxy
-    image: 'index.docker.io/sourcegraph/github-proxy:insiders@sha256:1ee9534f48d4855a429a2611bf1ebbfe40f3dd9d7a1db2281ae20dd64ee0e243'
+    image: 'index.docker.io/sourcegraph/github-proxy:insiders@sha256:5c9fc95300ef3296ba8c1d87b7477ebd2b272c6c552d834ab7ec762695d351bb'
     cpus: 1
     mem_limit: '1g'
     environment:
       - 'SRC_FRONTEND_INTERNAL=sourcegraph-frontend-internal:3090'
-      - JAEGER_AGENT_HOST=jaeger
+      - JAEGER_AGENT_HOST
```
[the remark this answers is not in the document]
bobheadxi: Oops this is an accident!

michaellzc:
> I don't think it's possible to make append environments with overrides, setting an override IIRC replaces environment completely

env var from different docker-compose override files can actually be stacked up together

this is how we override stuff for MI today https://github.com/sourcegraph/deploy-sourcegraph-managed/blob/main/admiral/red/docker-compose/docker-compose.override.yaml#L22

bobheadxi:
TIL, nice! Will make the updates.

bobheadxi:
Moved to override - though I realized I made an error in the initial implementation, so the current value requires https://github.com/sourcegraph/sourcegraph/pull/38910

docker-compose/otel/docker-compose.yaml
```diff
@@ -0,0 +1,65 @@
...
+    volumes:
+      - '../otel-collector/config.yaml:/etc/otel-collector/config.yaml'
+    cpus: 0.5
+    mem_limit: '512m'
+    ports:
+      - '0.0.0.0:4317:4317'
+      - '0.0.0.0:55679:55679'
```
michaellzc, line 12: is it necessary to bind these ports to the host?

I think each service will only send telemetry to `otel-collector:4317` so traffic won't go outside of the `sourcegraph` network

I assume `zpages` is some sort of debugging web interface? I think we can remove the port binding of `56679` but only include it in the `dev/docker-compose.yaml` override.

docker-compose/otel/docker-compose.yaml
```diff
@@ -0,0 +1,65 @@
+version: '2.4'
+services:
+  otel-collector:
```
michaellzc, line 3: it looks like we have included the health check extension

we can mimic the same probe in docker-compose https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_pod.tpl#L46-L52 to perform health check

docker-compose/otel/docker-compose.yaml
michaellzc: > will follow up to switch this to a sourcegraph image as we get closer to making this generally available

do you plan on doing this prior mergin this PR?

docker-compose/otel/docker-compose.yaml
bobheadxi: Note this mirrors how Jaeger binds _everything_ to the host: https://sourcegraph.com/github.com/sourcegraph/deploy-sourcegraph-docker/-/blob/docker-compose/docker-compose.yaml?L528-541

bobheadxi: That said, `zpages` is not really that useful either, so I will just remove these

bobheadxi: Removed in 0bebd7ebdd14b5c8e615c66371288f96f9818276, and tested these still work

docker-compose/otel/docker-compose.yaml
bobheadxi: This is kind of nontrivial actually, I think we need to build a custom collector image that can do the healthcheck without installing curl/wget (it's not included)

michaellzc: > This is kind of nontrivial actually, I think we need to build a custom collector image that can do the healthcheck without installing curl/wget (it's not included)

ughh right…

now I remember docker compose health check is a joke 🫤 (we tried to do something similar for cloud sql proxy in MI but turns out it’s impossible for the same reason)

bundling wget curl doesn’t feel right to me since I heard we’re trying to remove them from our images?

let’s save health check for k8s distro

michaellzc (approved):
nice!

docker-compose/otel/docker-compose.yaml
bobheadxi: > bundling wget curl doesn’t feel right to me since I heard we’re trying to remove them from our images?

Exactly what I was thinking - I tried to add something here but it didn't work: https://github.com/sourcegraph/sourcegraph/pull/39002/commits/e8bf8dff3186752cad502f0abc6760e02200c187 something to do with users that's beyond me, the binary is not available no matter how I spin it :lolsob:

I think I'll just leave healthchecks out of this for now

docker-compose/otel/docker-compose.yaml
bobheadxi: This will be available after https://github.com/sourcegraph/sourcegraph/pull/39002 , I'll follow up then

PR #839, opened 2022-07-20
otel: use sourcegraph image, detach config

bobheadxi:
### Checklist

### Test plan

same as https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/838


docker-compose/otel/docker-compose.yaml
```diff
@@ -7,17 +7,15 @@ services:
...
+    image: 'index.docker.io/sourcegraph/opentelemetry-collector:162246_2022-07-20_15dd0e0df1cc@sha256:3d64a9d20a791ba1c5d717ecb64cc19a10b1e30d2aac277143e773c5a9d28c2f'
     cpus: 0.5
     mem_limit: '512m'
     networks:
       - sourcegraph
     restart: always
     environment:
-      - JAEGER_AGENT_HOST=jaeger
-    command: ['--config', '/etc/otel-collector/config.yaml']
+      - JAEGER_AGENT=jaeger
```
michaellzc, line 17: would you point me to the docs where I can read more about this env var?

It looks like we're using `JAEGER_AGENT_HOST` everywhere else, is this one of the OTEL collector magic env var?

bobheadxi: https://github.com/sourcegraph/sourcegraph/pull/39002/files#diff-e1bc98a6e27a34fc34e6ef766282c0d1fea2634bb2720562cad280c5d0420c6aR5

bobheadxi: oops forgot to push a fix one sec, this should be JAEGER_HOST

bobheadxi: there's a slight discrepancy unfortunately but I think it makes sense in the context of the otel-collector as a standalone service - it just wants to be pointed at a jaeger instance

michaellzc (approved):

PR #857, opened 2022-09-22
Update tags for 4.0 release

camdencheek:
A recent change caused the `update-docker-tags.sh` script to only run on the `./pure-docker` subdirectory, but we also want to run it against the `./docker-compose` subdirectory. Will make a separate PR to update the script to run against both, but the is the result of running `go run ./tools/enforce-tags/ "4.0.0" ./docker-compose` manually.

I cherry-picked Coury's commit on `master` that updates the pinned version so we also update the otel images.

### Checklist

### Test plan

Manually verified that the tags in `docker-compose.yaml` are now populated correctly

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,6 @@ module github.com/sourcegraph/deploy-sourcegraph-docker
 go 1.15
 
 require (
-	github.com/sourcegraph/sourcegraph/enterprise/dev/ci/images v0.0.0-20220511200824-abaeea3fde03
+	github.com/sourcegraph/sourcegraph/enterprise/dev/ci/images v0.0.0-20220922183713-4ab710dc0025
 	github.com/sourcegraph/update-docker-tags v0.10.0
 )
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -3,6 +3,6 @@ module github.com/sourcegraph/deploy-sourcegraph-docker
 go 1.15
 
 require (
-	github.com/sourcegraph/sourcegraph/enterprise/dev/ci/images v0.0.0-20220511200824-abaeea3fde03
+	github.com/sourcegraph/sourcegraph/enterprise/dev/ci/images v0.0.0-20220922183713-4ab710dc0025
```
coury-clark, line 6: Did you mean to include this? I think this is the old pin?

camdencheek: I'm confused. I changed it to the new pin by cherry-picking your commit from `master`, right? This updates it from a version from May to a version from September

coury-clark: 🤦 sorry, misread the diff

coury-clark (approved):

PR #858, opened 2022-09-22
Update script to also run on ./docker-compose subdirectory

camdencheek:
It looks like this was changed in 6d8ecf323dfb7857 to only run on `./pure-docker` when that was moved to a subdirectory, but that causes the images tags in `./docker-compose` to no longer be updated by this script.

### Checklist

### Test plan

Ran this manually in https://github.com/sourcegraph/deploy-sourcegraph-docker/pull/857

### DIFF
--- a/tools/update-docker-tags.sh
+++ b/tools/update-docker-tags.sh
@@ -8,3 +8,4 @@ cd "$root_dir"
 CONSTRAINT=$1
 
 go run ./tools/enforce-tags "$CONSTRAINT" ./pure-docker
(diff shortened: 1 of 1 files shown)


coury-clark (approved):

michaellzc (approved):
@bobheadxi 

I thought the `sg ops update-images -kind compose` was supposed to replace the enforce-tags script?

bobheadxi:
> I thought the sg ops update-images -kind compose was supposed to replace the enforce-tags script?

IIRC it was only ever used for MI, it was never brought into the release tool

PR #859, opened 2022-09-23
prometheus: add zoekt-webserver scrape target

ggilmore:
for whatever reason, this wasn't in our prometheus configuration already

### Checklist

* [ ] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: Unnecessary since the k8s Prometheus configuration doesn't manually list scrape targets: https://github.com/sourcegraph/deploy-sourcegraph/blob/master/base/prometheus/prometheus.ConfigMap.yaml
### Test plan

Run docker-compose locally, navigate to the Prometheus instance `localhost:9090`, see that `zoekt-webserver` is listed as one of the targets under `localhost:9090/targets`

### DIFF
--- a/prometheus/prometheus_targets.yml
+++ b/prometheus/prometheus_targets.yml
@@ -34,6 +34,10 @@
   targets:
     - zoekt-indexserver-0:6072
 - labels:
+    job: zoekt-webserver
+  targets:
+    - zoekt-webserver-0:6070
+- labels:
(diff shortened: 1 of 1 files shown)


abeatrix (approved):
👍

stefanhengl (approved):

keegancsmith (approved):

PR #877, opened 2022-11-09
prometheus: change scrape interval to 30s

ggilmore:
This change docker-compose's Prometheus scrape interval to be 30s, which brings it inline with our existing k8s configuration: 

https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/22bed12869be633318d7b56cf3ad8f04d9b4b67b/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml#L5-L7

Doing so allows us to create "higher resolution" dashboards that aggregate over 1m (~twice the sample rate, following [Nyquist–Shannon](https://www.allaboutcircuits.com/technical-articles/nyquist-shannon-theorem-understanding-sampled-systems/)).

### Checklist

* [x] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: N/A
### Test plan

Ran local docker-compose dev stack,  verified that prometheus dashboards still work:

Screen Shot 2022-11-09 at 9 31 13 AM

### DIFF
--- a/prometheus/docker_rules.yml
+++ b/prometheus/docker_rules.yml
@@ -1,9 +1,5 @@
 # Docker container monitoring rules
 
-global:
-  scrape_interval:     30s
-  evaluation_interval: 30s
-
 groups:
   - name: docker.rules
     rules:
--- a/prometheus/docker_rules.yml
+++ b/prometheus/docker_rules.yml
@@ -1,5 +1,9 @@
 # Docker container monitoring rules
 
+global:
+  scrape_interval:     30s
+  evaluation_interval: 30s
+
 groups:
   - name: docker.rules
     rules:
--- a/prometheus/prometheus.yml
+++ b/prometheus/prometheus.yml
@@ -1,0 +1,35 @@
+# Prometheus global config
+global:
+  scrape_interval:     30s
+  evaluation_interval: 30s
+  # scrape_timeout is set to the global default (10s).
+
+# Alertmanager configuration
+alerting:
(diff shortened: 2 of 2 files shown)


jhchabran (approved):

jac:
@ggilmore any reason to make the change here versus in the [config](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/docker-images/prometheus/config/prometheus.yml) that gets baked into the Prometheus Docker image we distribute?

ggilmore:
We already already override that configuration in our Kubernetes deployment target (we need custom tweaks for k8s, etc.): see https://github.com/sourcegraph/deploy-sourcegraph-helm/blob/main/charts/sourcegraph/templates/prometheus/prometheus.ConfigMap.yaml (we override prometheus.yml) there too. 

Doing it this way also makes it clearer for customers how'd they'd tweak this configuration if they want to do so.

jac:
Ah that makes sense, thanks!

bobheadxi:
@ggilmore can you update the bundled configuration then? It still has `1m` as the scrape interval https://github.com/sourcegraph/sourcegraph/blob/main/docker-images/prometheus/config/prometheus.yml

bobheadxi:
Additionally can you document that each deployment method's prom config should be aligned? Preferably in the docstring of each of these configs

bobheadxi:
https://github.com/sourcegraph/sourcegraph/pull/44168

PR #878, opened 2022-11-15
Move docker compose installation scripts from docsite

abeatrix:
Related to https://github.com/sourcegraph/sourcegraph/pull/44449

Currently, the docker-compose installation scripts for each cloud provider are stored in our doc site, and it takes up half of the page, [example](https://docs.sourcegraph.com/admin/deploy/docker-compose/aws#advanced-details-user-data):
image

This PR proposed to move the scripts from doc site to the [docker compose reference repo](https://github.com/sourcegraph/deploy-sourcegraph-docker), and allow users to install via curl command instead of copying a big block of text. Example:
image

Same format is currently being used in our k3s installation guide: https://docs.sourcegraph.com/admin/deploy/single-node/k3s#start-up-script

### Checklist

* [N/A] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change:
* [N/A] All images have a valid tag and SHA256 sum

### Test plan

These are the same scripts that are being used by our customers from our docs site.

### DIFF
--- a/docker-compose/scripts/install-aws.sh
+++ b/docker-compose/scripts/install-aws.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+###############################################################################
+# ACTION REQUIRED: REPLACE THE URL AND REVISION WITH YOUR DEPLOYMENT REPO INFO
+###############################################################################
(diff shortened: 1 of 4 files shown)


jdpleiness (approved):
👍

PR #890, opened 2022-12-27
Add relative path for rules in sg_config_prometheus

gabtorre:
After upgrading to 4.2, the alerts were missing in Prometheus.

Screen Shot 2022-12-26 at 2 47 24 PM

This is the config in Prometheus. The root directory of [prometheus.yml](https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/prometheus/prometheus.yml) is now [`sg_prometheus_add_ons`](https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/master/docker-compose/docker-compose.yaml#L463), so there is no path to the rules in `/sg_config_prometheus/`
https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/docker-images/prometheus/prometheus.sh?L7 

Screen Shot 2022-12-26 at 2 48 03 PM

Adding the path to the rules in `/sg_config_prometheus` added `/sg_config_prometheus/` to the config and brought the alerts back. This is how it's configured in k8s too: 
https://github.com/sourcegraph/deploy-sourcegraph/blob/master/base/prometheus/prometheus.ConfigMap.yaml#L26

Screen Shot 2022-12-26 at 3 23 45 PM

Screen Shot 2022-12-26 at 3 24 05 PM

### Checklist

* [ ] Sister [deploy-sourcegraph](https://github.com/sourcegraph/deploy-sourcegraph) change: Only affects Docker deployments
### Test plan
Checked that the alerts were present in Prometheus and Grafana.

### DIFF
--- a/prometheus/prometheus.yml
+++ b/prometheus/prometheus.yml
@@ -15,7 +15,7 @@ alerting:
 
 # Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
 rule_files:
-  - '*_rules.yml'
+  - '/sg_config_prometheus/*_rules.yml'
   - '/sg_prometheus_add_ons/*_rules.yml'
 
 # Configure targets to scrape


abeatrix (approved):
Nice catch!

PR #1204, opened 2025-05-09
[REL-961] Close Grafana port from outside access

marcleblanc2:
[REL-961 Review Docker Compose file for Grafana port exposure security](https://linear.app/sourcegraph/issue/REL-961/review-docker-compose-file-for-grafana-port-exposure-security)

Is there any particular reason why we have the Grafana port 3370 open for side channel access? It seems like it's been open since Grafana was first bundled with the product in https://github.com/sourcegraph/deploy-sourcegraph-docker/commit/51ca07933b2149f3fb038b8969d95d4be8843756, even though the reverse proxy was implemented at about the same time, in https://github.com/sourcegraph/sourcegraph/commit/939b612fb319f6563116f4bcd5814dc486f78361

I suggest we close the side channel access, to reduce exposure for customers running the Airgapped Analytics dashboard. 

Also, fixed a broken port number in sourcegraph-frontend-internal's config for the Grafana URL, which has been broken since https://github.com/sourcegraph/deploy-sourcegraph-docker/commit/f6f8d8dff29335a4a96133b58edb0d931415fca9, so probably not used. 

### Checklist

* [ ] Sister [customer-replica](https://github.com/sourcegraph/deploy-sourcegraph-docker-customer-replica-1) change (if necessary, for any changes affecting pure-docker or configuration):
### Test plan

Tested on test instance, frontend still proxies the connection to Grafana as needed, without side channel access

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -192,7 +192,7 @@ services:
       - 'SYMBOLS_URL=http://symbols-0:3184'
       - 'INDEXED_SEARCH_SERVERS=zoekt-webserver-0:6070'
       - 'SRC_FRONTEND_INTERNAL=sourcegraph-frontend-internal:3090'
-      - 'GRAFANA_SERVER_URL=http://grafana:3000'
+      - 'GRAFANA_SERVER_URL=http://grafana:3370'
       - 'PROMETHEUS_URL=http://prometheus:9090'
       - 'PRECISE_CODE_INTEL_UPLOAD_BACKEND=blobstore'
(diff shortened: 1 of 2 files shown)


DaedalusG (approved):
I did a review and this looks good to me, I haven't tried a deploy of these changes myself, but I trust your test. I actually think this is a better design in general 👍🏼 

Might want to run it by security but I think having people have to auth through the frontend to look at the grafana makes more sense than leaving exposed as a side car on the host machine. 

Final thought is just that hopefully we dont have any on prem people who have some odd configuration relying on this. Might be worth a note in the deployment type upgrade notes for docker-compose

PR #1252, opened 2025-07-22
Add checkov CI check

willdollman:
Add checkov CI step - we require a SAST check on key repos for SOC 2 compliance.

## Test plan
Confirm checkov check passes


evict:
We already have this in buildkite: https://github.com/sourcegraph/deploy-sourcegraph-docker/blob/main/.buildkite/ci-checkov.sh

PR #1420, opened 2026-09-04
fix/docs: Fix spelling errors in the documentation

marcleblanc2:
Corrections were produced by running [codespell](https://github.com/codespell-project/codespell) over the repository's hand-written Markdown, then reviewing every proposed change in context so that code samples, CLI flags, config keys, YAML keys and values, URLs, identifiers, and quoted output were left untouched. Only spelling is changed; no wording, formatting, or content was otherwise altered.

Files that are generated or vendored are skipped, since fixes there would be overwritten. Where a generated file has a checked-in source, the source is corrected and the generated output is regenerated so the two stay in sync.

[_Created by a Sourcegraph agentic batch change._](https://sourcegraph.sourcegraph.com/batch-change-agents/1136)

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -2,12 +2,12 @@
 
 Developers can test using [Vagrant](https://www.vagrantup.com) and your Sourcegraph GCP account.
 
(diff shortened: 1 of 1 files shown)


filiphaftek (approved):
TY

